PR #281, opened 2022-02-25
[P2PW-1144] - Fully working send feature

1sakov:


solanaj/src/main/java/org/p2p/solanaj/kits/TokenTransaction.kt
```diff
@@ -52,7 +52,7 @@ object TokenTransaction {
         return AccountInfoData.decode(data)
     }
 
-    fun decodeAccountInfo(value: AccountInfo.Value): AccountInfoData {
+    fun decodeAccountInfo(@NonNull value: AccountInfo.Value): AccountInfoData {
```
kabramovp2p, line 55: @NonNull? =) in kotlin?

kabramovp2p (approved):

solanaj/src/main/java/org/p2p/solanaj/kits/TokenTransaction.kt
1sakov: Глаза мазолятся)

PR #349, opened 2022-03-14
[P2PW-1482] Search token error

davran1001:


app/src/main/java/org/p2p/wallet/receive/list/TokenListData.kt
```diff
@@ -0,0 +1,10 @@
+package org.p2p.wallet.receive.list
+
+import org.p2p.wallet.user.model.TokenData
+
+data class TokenListData(
+    val searchText: String = "",
+    val result: List<TokenData> = emptyList()
+) {
+    fun getSize() = result.size
```
kabramovp2p, line 9: maybe not fun?
just 
```
   val size = result.size
```

kabramovp2p (approved):

gslevinkov:
you haven't put your self in the assignee tab

app/src/main/java/org/p2p/wallet/receive/list/TokenListData.kt
```diff
@@ -0,0 +1,10 @@
+package org.p2p.wallet.receive.list
+
+import org.p2p.wallet.user.model.TokenData
+
+data class TokenListData(
```
gslevinkov, line 5: maybe you should add some context that this class is data for the screen
smth like `TokenListScreenDetails` `TokenListScreenData`
and im not sure that default parameters is a good choice here

app/src/main/java/org/p2p/wallet/receive/list/TokenListFragment.kt
```diff
@@ -75,6 +75,18 @@ class TokenListFragment :
...
                 }
             }
+            recyclerView.isVisible = items.isNotEmpty()
+            emptyView.isVisible = items.isEmpty()
+        }
+    }
+
+    override fun showEmpty(searchText: String) {
+        with(binding) {
+            val text = getString(R.string.receive_token_list_not_found, searchText)
```
gslevinkov, line 85: yeah, I know that getString returns `text` :) 
what im interested in when reading this code - what kind of text? 
`notFoundText` `notFoundTitle` is good enough IMHO

app/src/main/java/org/p2p/wallet/receive/list/TokenListPresenter.kt
```diff
@@ -48,8 +47,12 @@ class TokenListPresenter(
 
     private fun observeTokens() {
         launch {
-            interactor.getTokenListFlow().collect { tokens ->
-                view?.showItems(tokens, scrollToUp)
+            interactor.getTokenListFlow().collect { data ->
+                if (data.result.isNotEmpty()) {
+                    view?.showItems(data.result, scrollToUp)
```
gslevinkov, line 52: have you tried resolve such things using sealed ViewState class? only one func in view - renderViewState

app/src/main/res/layout/fragment_receive_list.xml
```diff
@@ -81,7 +81,45 @@
...
             tools:listitem="@layout/item_token_list" />
+
+        <androidx.constraintlayout.widget.ConstraintLayout
+            android:id="@+id/emptyView"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:layout_margin="16dp"
+            android:background="@drawable/bg_system_error_main_rounded"
+            android:padding="16dp">
+
```
gslevinkov, line 94: needless line breaks

app/src/main/java/org/p2p/wallet/user/repository/UserInMemoryRepository.kt
```diff
@@ -3,60 +3,93 @@ package org.p2p.wallet.user.repository
...
+                setResult(searchText)
             }
-            val indexes = findEdges(items = result, count = count, offset = offset)
-            appendTokens(result.subList(indexes.first, indexes.second))
         }
+    }
 
-        val indexes = findEdges(items = items, count = count, offset = offset)
-        val result = items.subList(indexes.first, indexes.second)
+    private fun setResult(key: String) {
```
gslevinkov, line 70: so, are you setting here only DEFAULT result? or search result?
i can see you are using only DEFAULT_TOKEN_KEY

app/src/main/java/org/p2p/wallet/user/repository/UserInMemoryRepository.kt
```diff
@@ -3,60 +3,93 @@ package org.p2p.wallet.user.repository
...
+                val newSize = (defaultTokens.size - getPopularDecimals().size) + count
+                val needToLoadMore = newSize > defaultTokens.size
+                if (needToLoadMore) {
+                    searchTextByTokens[DEFAULT_TOKEN_KEY] = getPopularDecimals() + allInMemoryTokens.take(newSize)
+                }
+                setResult(DEFAULT_TOKEN_KEY)
+            }
+            searchText in searchTextByTokens -> {
+                val tokensBySearchText = searchTextByTokens.getOrDefault(searchText, emptyList())
+                val newSize = tokensBySearchText.size + count
```
gslevinkov, line 56: maybe `count` is not a good choice of words here
what count really means? 
it's like new search list size, right? maybe we can think in this direction and name it better

app/src/main/java/org/p2p/wallet/user/repository/UserInMemoryRepository.kt
```diff
@@ -3,60 +3,93 @@ package org.p2p.wallet.user.repository
...
-                    it.name.startsWith(searchText, ignoreCase = true)
+        when {
+            searchText.isEmpty() -> {
+                val defaultTokens = searchTextByTokens.getOrDefault(DEFAULT_TOKEN_KEY, emptyList())
+                val newSize = (defaultTokens.size - getPopularDecimals().size) + count
+                val needToLoadMore = newSize > defaultTokens.size
+                if (needToLoadMore) {
+                    searchTextByTokens[DEFAULT_TOKEN_KEY] = getPopularDecimals() + allInMemoryTokens.take(newSize)
+                }
+                setResult(DEFAULT_TOKEN_KEY)
```
gslevinkov, line 52: how bout DEFAULT_TOKEN_KEY -> EMPTY_SEARCH_TEXT_KEY?
cause we are using this key when searchText is empty

gslevinkov (changes requested):

PR #351, opened 2022-03-14
P2PW-1260 - Change scale type of buy amount and change button text

davran1001:


app/src/main/java/org/p2p/wallet/renbtc/ui/info/RenBtcBuyBottomSheet.kt
```diff
@@ -53,8 +53,14 @@ class RenBtcBuyBottomSheet : NonDraggableBottomSheetDialogFragment() {
         with(binding) {
             val feeUsd = if (priceInUsd != null) "~$$priceInUsd" else getString(R.string.common_not_available)
             topTextView.text = getString(R.string.send_account_creation_fee_format, feeUsd)
-            amountTextView.text = priceInSol.toString()
+            amountTextView.text = getString(R.string.receive_amount_in_sol, priceInSol.toString())
+            val actionText = getString(R.string.receive_pay_and_countinue, priceInSol.toString())
```
gslevinkov, line 57: можно сразу поправить countinue на continue и заinlineнить вызов прям в setActionText

gslevinkov (approved):

kuznetsov (approved):

1sakov (approved):

app/src/main/java/org/p2p/wallet/rpc/interactor/TokenInteractor.kt
```diff
@@ -0,0 +1,73 @@
...
+import org.p2p.wallet.rpc.repository.RpcRepository
+import org.p2p.wallet.utils.toPublicKey
+
+class TokenInteractor(
+    private val rpcRepository: RpcRepository,
+    private val addressInteractor: TransactionAddressInteractor,
+    private val tokenKeyProvider: TokenKeyProvider
+) {
+
+    suspend fun close(addressToClose: String): String {
```
gslevinkov, line 16: from the current naming I can't understand what we are closing really

app/src/main/java/org/p2p/wallet/rpc/interactor/TokenInteractor.kt
```diff
@@ -0,0 +1,73 @@
...
+        val owner = tokenKeyProvider.publicKey.toPublicKey()
+        val instruction = TokenProgram.closeAccountInstruction(
+            TokenProgram.PROGRAM_ID,
+            addressToClose.toPublicKey(),
+            owner,
```
gslevinkov, line 21: named arg is better here

app/src/main/java/org/p2p/wallet/rpc/interactor/TokenInteractor.kt
```diff
@@ -0,0 +1,73 @@
...
+            owner
+        )
+
+        val transaction = Transaction()
+        transaction.addInstruction(instruction)
```
gslevinkov, line 26: can be rewritten with `.apply` that is used exactly for initialising objects

app/src/main/java/org/p2p/wallet/rpc/interactor/TokenInteractor.kt
```diff
@@ -0,0 +1,73 @@
...
+        val recentBlockhash = rpcRepository.getRecentBlockhash()
+        transaction.recentBlockHash = recentBlockhash.recentBlockhash
+
+        val signers = Account(tokenKeyProvider.secretKey)
+        transaction.sign(signers)
+
+        return rpcRepository.sendTransaction(transaction)
+    }
+
+    suspend fun open(
```
gslevinkov, line 37: The same poing with naming here. What are we opening?

app/src/main/java/org/p2p/wallet/rpc/interactor/TokenInteractor.kt
```diff
@@ -0,0 +1,73 @@
...
+
+        val account = Account(tokenKeyProvider.secretKey)
+        val feePayer = account.publicKey
+
+        val splDestinationAddress = addressInteractor.findSplTokenAddressData(
+            destinationAddress = tokenKeyProvider.publicKey.toPublicKey(),
+            mintAddress = mintAddress
+        )
+
+        // get address
```
gslevinkov, line 49: better to extract this logic to method `getAddress` and remove comment

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/rpc/interactor/TokenInteractor.kt
gslevinkov: For example I can inject your interactor inside my class and use it this way:
```
val interactor : TokenInteractor by inject()

interactor.open(....)
interactor.close(...)
```
it looks like we are closing and opening the interactor, but not the token itself

gslevinkov (changes requested):

app/src/main/java/org/p2p/wallet/rpc/interactor/TokenInteractor.kt
davran1001: it's java class which does not support named args

gslevinkov (approved):

PR #355, opened 2022-03-14
P2PW-1534 - Refactor setStatusBarColor and setStatusBarIconColors, re…

davran1001:
…move deprecate methods and fix navigation bar icons color


1sakov (approved):

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/common/mvp/BaseFragment.kt
```diff
@@ -70,20 +73,26 @@ abstract class BaseFragment(@LayoutRes layoutRes: Int) : Fragment(layoutRes), Ba
...
-            decor.systemUiVisibility = 0
-        }
+    /*
+        Change status bar icon color
+        isLight = true - dark color
+        isLight = false - white color
+
+        Don't forget to reset status bar color on fragment destroy to restore previous color
+     */
+    protected fun setLightStatusBar(isLight: Boolean) {
```
gslevinkov, line 92: maybe we should find another name than `isLight` that removes kdoc need?
or are we changing status bar icon color? so maybe we should rename a whole method?
`setStatusBarIconsList(makeIconsLight: Boolean)`

app/src/main/java/org/p2p/wallet/receive/renbtc/ReceiveRenBtcFragment.kt
```diff
@@ -36,14 +36,14 @@ class ReceiveRenBtcFragment :
...
     override val statusBarColor: Int = R.color.backgroundButtonPrimary
-    override val statusBarDarkTint: Boolean = false
 
     override val presenter: ReceiveRenBtcContract.Presenter by inject()
     private val binding: FragmentRenBtcBinding by viewBinding()
     private val analyticsInteractor: AnalyticsInteractor by inject()
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
+        setLightStatusBar(false)
```
gslevinkov, line 46: named arg is better when working with boolean args

gslevinkov (changes requested):

app/src/main/java/org/p2p/wallet/common/mvp/BaseFragment.kt
gslevinkov: it would be better to make kdoc this way:
```
    /*
        Change status bar and its icons color
        isLight = true - white status bar, dark icons
        isLight = false - dark status bar, white icons
        Don't forget to reset status bar color on fragment destroy to restore previous color
     */
```

gslevinkov (approved):

PR #357, opened 2022-03-15
P2PW-1412 - fix amounts on Send Screen

kabramovp2p:


kabramovp2p:
Android Emulator - Pixel_5_API_30:5554 2022-03-15 12-47-36

app/src/main/java/org/p2p/wallet/home/model/Token.kt
```diff
@@ -71,9 +71,9 @@ sealed class Token constructor(
 
         fun getFormattedTotal(includeSymbol: Boolean = false): String =
             if (includeSymbol) {
-                "${total.scaleLong().toPlainString()} $tokenSymbol"
+                "${AmountUtils.format(total)} $tokenSymbol"
             } else {
-                total.scaleLong().toPlainString()
+                AmountUtils.format(total)
```
gslevinkov, line 76: are we sure that we want to sticks with Util ending plus using static methods like in Java?
maybe extension would be better here?
`total.formatWithSeparator()`

kabramovp2p: i agree to make it as an extension but we need to rise this question to @1sakov

gslevinkov: im okay to create tech-task for this, or fix right in this task

kabramovp2p: im also do not care when and where to fix
right now or in other tasks =)

1sakov: But it will create new formatter object every time. 
We need avoid it and use the already created formatter, don't we?

gslevinkov: but it will be wiped out by garbage collector after we pop out from the func

I guess we need to find a trade-off. 
1. Make it an ext, but formatter is a static field that lives while the app lives
2. Make it a class, but inject it everywhere where formatting is needed
3. ??

1sakov: Why the formatted in `AmountUtils` will be popped? 
It's the same as `Make it an ext, but formatter is a static field that lives while the app lives`, but the fun is not extension

gslevinkov (approved):

1sakov (approved):

kuznetsov (approved):

PR #366, opened 2022-03-18
P2PW-1582 transaction limits dialog added on OrcaSwapFragment.kt

kabramovp2p:


app/src/main/java/org/p2p/wallet/swap/interactor/orca/OrcaSwapInteractor.kt
```diff
@@ -61,6 +61,12 @@ class OrcaSwapInteractor(
         feePayerToken = newToken
     }
 
+    suspend fun getFreeTransactionsInfo(): Pair<Int, Int> {
```
gslevinkov, line 64: let's try and remove Pair here? Pair is not readable at all, it has primitives in it, and only two methods `first`, `second`

I suggest you replace the Pair with simple class

kabramovp2p: i do not like not needed classes 
if we can use 2 params i prefer use Pair
if there is more then 2.. then class

gslevinkov: Okay
Pair is good when named good. 
I don't know what to expect when I call `getFreeTransactionsInfo`. Maybe at least we can rename the method to tell us what we are expecting as a result?

getFreeTransactionFeeMaxUsageToFeeRemaining

it's like I don't know what type of int lies in Pair. I need to check all the implementation inside to understand what methods returns
```
val intOne = pair.first 
val intTwo = pair.second
```

gslevinkov (approved):

davran1001 (approved):

PR #368, opened 2022-03-18
P2PW-1263 - Token swipes fixes

kuznetsov:


kuznetsov:
I think we should get rid of separate item and ViewHolder for Hidden Token, cause it's almost fully duplicate the Token code

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/common/ui/recycler/SwipeLayout.java
```diff
@@ -633,6 +639,7 @@ private void updateState() {
     private GestureDetector.OnGestureListener gestureDetectorCallBack = new GestureDetector.SimpleOnGestureListener() {
         @Override
         public boolean onScroll(MotionEvent e1, MotionEvent e2, float distanceX, float distanceY) {
+            // here
```
gslevinkov, line 642: `here`? :)

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/common/ui/recycler/SwipeLayout.java
kuznetsov: Ahah, thanks :)

PR #371, opened 2022-03-21
P2PW-1635 - ProgressBottomSheet buttons redesign

kabramovp2p:
P2PW-1635 - remove two buttons -> make only one "Done" and reuse common logic


app/src/main/java/org/p2p/wallet/transaction/ui/ProgressBottomSheet.kt
```diff
@@ -187,4 +155,20 @@ class ProgressBottomSheet : NonDraggableBottomSheetDialogFragment() {
...
+            titleTextView.text = state.message
+            progressBar.isVisible = false
+
+            transactionImageView.setImageResource(R.drawable.ic_error_transaction)
+            lineView.isVisible = true
+            lineView.setBackgroundColor(getColor(R.color.systemErrorMain))
+        }
+    }
+
+    private fun setResultAndDismiss(result: Bundle = bundleOf(EXTRA_RESULT_KEY_DISMISS to null)) {
```
gslevinkov, line 170: the bundle is always `EXTRA_RESULT_KEY_DISMISS to null`, check your method calls! 
or remove this parameter

gslevinkov (changes requested):

gslevinkov (approved):

app/src/main/res/values/strings.xml
```diff
@@ -433,8 +433,7 @@
     <string name="details_value">Value</string>
     <string name="details_paid_by_p2p">Paid by p2p.org</string>
     <string name="details_paid_by_p2p_brackets">(Paid by p2p.org)</string>
-    <string name="details_show_transaction_details">Show transaction details</string>
-    <string name="details_hide_transaction_details">Hide transaction details</string>
+    <string name="details_done">@string/common_done</string>
```
1sakov, line 436: seems it's redundant

app/src/main/java/org/p2p/wallet/transaction/ui/ProgressBottomSheet.kt
```diff
@@ -122,55 +113,32 @@ class ProgressBottomSheet : NonDraggableBottomSheetDialogFragment() {
...
             transactionIdGroup.isVisible = true
-            transactionIdTextView.text = state.transaction.signature
+            transactionIdTextView.text = signature
 
             transactionImageView.setImageResource(R.drawable.ic_success)
             lineView.isVisible = true
             lineView.setBackgroundColor(getColor(R.color.systemSuccessMain))
 
-            showButton.setOnClickListener {
-                val result = bundleOf(EXTRA_RESULT_KEY_PRIMARY to state.transaction)
```
1sakov, line 171: Let's remove this and secondary constant since it's not used anymore.
And you'll need to update callback listeners in Send and Swap screens as well

app/src/main/res/values/strings.xml
kabramovp2p: why? :(

kabramovp2p: ohh you mean use common_done directly?

app/src/main/java/org/p2p/wallet/transaction/ui/ProgressBottomSheet.kt
kabramovp2p: fixed

app/src/main/res/values/strings.xml
1sakov: yeah, exactly

1sakov (approved):

PR #374, opened 2022-03-22
P2PW-1633 - add searchView in SelectTokenFragment

davran1001:
Add support for search tokens in SearchTokenFragment


app/src/main/java/org/p2p/wallet/home/ui/select/SelectTokenFragment.kt
```diff
@@ -62,4 +67,32 @@ class SelectTokenFragment : BaseFragment(R.layout.fragment_select_token) {
             emptyTextView.isVisible = isEmpty
         }
     }
+
+    override fun onQueryTextSubmit(query: String?): Boolean = false
+
+    override fun onQueryTextChange(newText: String?): Boolean {
+        val searchText = newText.orEmpty()
```
gslevinkov, line 74: it looks like that all this filtering and checks should be in presenter, it's his responsibility to define what to show on UI

app/src/main/java/org/p2p/wallet/home/ui/select/SelectTokenFragment.kt
```diff
@@ -62,4 +67,32 @@ class SelectTokenFragment : BaseFragment(R.layout.fragment_select_token) {
...
+        if (searchText.length < QUERY_MIN_LENGTH) {
+            tokenAdapter.setItems(tokens)
```
gslevinkov, line 76: now when we have another type of list of tokens, it's better to name this field as `allTokens`

app/src/main/java/org/p2p/wallet/home/ui/select/SelectTokenFragment.kt
```diff
@@ -62,4 +67,32 @@ class SelectTokenFragment : BaseFragment(R.layout.fragment_select_token) {
...
+            return true
+        }
+        val filteredItems = tokens.filter {
+            it.tokenName.startsWith(searchText, ignoreCase = true) || it.tokenSymbol == newText
```
kabramovp2p, line 80: i think we need to check .contains
tokenName and tokenSymbol .contains newText
to make search more UserFriendly
for now it.tokenSymbol == newText means that only if user will input USDC ... then token will appear

kabramovp2p, line 80: i think we need to check .contains
tokenName and tokenSymbol .contains newText
to make search more UserFriendly
for now it.tokenSymbol == newText means that only if user will input USDC ... then token will appear

app/src/main/res/menu/menu_search.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto">
+    <item
+        android:id="@+id/menu_search"
+        android:title="@string/search_menu"
+        app:actionLayout = "@layout/item_search_bar"
+        app:showAsAction="always" />
+</menu>
```
kabramovp2p, line 10: new line

PR #377, opened 2022-03-23
P2PW-1589 - Adaptive amount text size in send and swap

kuznetsov:


gslevinkov:
wrong commit naming

app/src/main/java/org/p2p/wallet/swap/ui/orca/OrcaSwapFragment.kt
```diff
@@ -77,40 +81,14 @@ class OrcaSwapFragment :
     private var onBackPressedCallback: OnBackPressedCallback? = null
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        Timber.d("onViewCreated()")
```
gslevinkov, line 84: no need for this I think

app/src/main/java/org/p2p/wallet/swap/ui/orca/OrcaSwapFragment.kt
```diff
@@ -149,6 +127,51 @@ class OrcaSwapFragment :
         presenter.loadInitialData()
     }
 
+    private fun setupViews() = with(binding) {
```
gslevinkov, line 130: thx for extracting all this to a separate method!

app/src/main/java/org/p2p/wallet/swap/ui/orca/OrcaSwapFragment.kt
```diff
@@ -149,6 +127,51 @@ class OrcaSwapFragment :
...
+        }
+        sourceImageView.setOnClickListener { presenter.loadTokensForSourceSelection() }
+        destinationImageView.setOnClickListener { presenter.loadTokensForDestinationSelection() }
+        destinationTextView.setOnClickListener { presenter.loadTokensForDestinationSelection() }
+        availableTextView.setOnClickListener { presenter.calculateAvailableAmount() }
+        maxTextView.setOnClickListener { presenter.calculateAvailableAmount() }
+
+        setupAmountListener()
+
+        exchangeImageView.setOnClickListener { presenter.reverseTokens() }
```
gslevinkov, line 147: maybe format all setOn*Listener in one style?

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/swap/ui/orca/OrcaSwapFragment.kt
kuznetsov: We can consider it

app/src/main/java/org/p2p/wallet/swap/ui/orca/OrcaSwapFragment.kt
kuznetsov: Tanks

kabramovp2p (approved):

1sakov (approved):

PR #378, opened 2022-03-23
PWN-3074 - Fix crash in reset seed phrase screen

kuznetsov:


app/src/main/java/org/p2p/wallet/settings/ui/reset/seedphrase/ResetSeedPhraseFragment.kt
```diff
@@ -84,7 +85,11 @@ class ResetSeedPhraseFragment :
                 keysRecyclerView.isVisible = true
                 phraseAdapter.addSecretKey(SecretKey())
             }
-            messageTextView.text = buildSeedInfoText()
+
+            messageTextView.apply {
```
gslevinkov, line 89: scope function is used when there are more than 2 calls

app/src/main/java/org/p2p/wallet/settings/ui/reset/seedphrase/ResetSeedPhraseFragment.kt
```diff
@@ -111,7 +116,14 @@ class ResetSeedPhraseFragment :
     }
 
     private fun buildSeedInfoText(): SpannableString {
-        val message = getString(R.string.auth_recover_info)
+        val seedInfo = getString(R.string.settings_what_is_a_security_key)
+
+        val message = StringBuilder().apply {
```
gslevinkov, line 121: can be replaced with `buildString`

1sakov (approved):
Approved from my side.
Didn't get what was the reason of the crash? 
It didn't find the span?

kabramovp2p (approved):

kuznetsov:
> Approved from my side. Didn't get what was the reason of the crash? It didn't find the span?

Spanned text was missing in general message, that led to exception in `span.setSpan()`

PR #381, opened 2022-03-23
PWN-3038 - Fixed fee relayer transactions for swap/send

1sakov:


kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/feerelayer/model/TokenInfo.kt
```diff
@@ -1,6 +1,12 @@
...
+import org.p2p.wallet.utils.Constants.WRAPPED_SOL_MINT
+
 class TokenInfo(
     val address: String,
     val mint: String
-)
+) {
+
+    val isSOL: Boolean
+        get() = WRAPPED_SOL_MINT == mint
```
gslevinkov, line 11: isSol can be a field. class and its fields are final, so it's okay to initialise value on creating instance

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/feerelayer/model/TokenInfo.kt
1sakov: That's true, will do

PR #387, opened 2022-03-24
PWN-3079 - hide devNet on builds for prod

kabramovp2p:
https://p2pvalidator.atlassian.net/browse/PWN-3079

Added AppSettings and use featureToggler for isProd to hide or show devnet onProdEnvs

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -44,9 +44,9 @@ android {
             buildConfigField("String", "intercomAppId", "\"\"")
             buildConfigField("String", "amplitudeKey", "\"\"")
         }
-        buildConfigField("boolean", "CRASHLYTICS_ENABLED", "false")
+        buildConfigField("boolean", "CRASHLYTICS_ENABLED", "true")
         buildConfigField("boolean", "ANALYTICS_ENABLED", "false")
-        buildConfigField("boolean", "IS_PROD", "false")
+        buildConfigField("boolean", "KEY_DEV_NET_ENABLED", "true")
         buildConfigField("String", "TASK_NUMBER", "\"${taskNumber()}\"")
 
         javaCompileOptions {
@@ -80,6 +80,7 @@ android {
             minifyEnabled false
             shrinkResources false
             versionNameSuffix ".$VERSION_BUILD-debug"
+            buildConfigField("boolean", "CRASHLYTICS_ENABLED", "false")
         }
 
         feature {
@@ -103,9 +104,8 @@ android {
             shrinkResources true
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
 
-            buildConfigField("boolean", "CRASHLYTICS_ENABLED", "true")
             buildConfigField("boolean", "ANALYTICS_ENABLED", "true")
-            buildConfigField("boolean", "IS_PROD", "true")
+            buildConfigField("boolean", "KEY_DEV_NET_ENABLED", "false")
         }
     }
 
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -44,7 +44,9 @@ android {
             buildConfigField("String", "intercomAppId", "\"\"")
             buildConfigField("String", "amplitudeKey", "\"\"")
         }
-
+        buildConfigField("boolean", "CRASHLYTICS_ENABLED", "false")
+        buildConfigField("boolean", "ANALYTICS_ENABLED", "false")
+        buildConfigField("boolean", "IS_PROD", "false")
         buildConfigField("String", "TASK_NUMBER", "\"${taskNumber()}\"")
 
         javaCompileOptions {
@@ -78,9 +80,6 @@ android {
             minifyEnabled false
             shrinkResources false
             versionNameSuffix ".$VERSION_BUILD-debug"
-            buildConfigField("boolean", "CRASHLYTICS_ENABLED", "false")
-            buildConfigField("boolean", "ANALYTICS_ENABLED", "false")
-            buildConfigField("boolean", "IS_PROD", "false")
         }
 
         feature {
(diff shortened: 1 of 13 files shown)


app/build.gradle
```diff
@@ -99,6 +97,7 @@ android {
 
             buildConfigField("boolean", "CRASHLYTICS_ENABLED", "true")
             buildConfigField("boolean", "ANALYTICS_ENABLED", "true")
+            buildConfigField("boolean", "IS_PROD", "true")
```
1sakov, line 100: But we can use `!BuildConfig.DEBUG` to determine if it's not release build, right? Isn't it redundant?

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/common/AppSettings.kt
```diff
@@ -0,0 +1,25 @@
...
+import android.content.SharedPreferences
+import androidx.core.content.edit
+import org.p2p.wallet.BuildConfig
+
+private const val KEY_POLLING_ENABLED = "KEY_POLLING_ENABLED"
+private const val KEY_IS_PROD = "KEY_IS_PROD"
+
+class AppSettings(private val sharedPreferences: SharedPreferences) {
+
+    val isPoolingEnabled: Boolean
```
1sakov, line 12: `isPollingEnabled`

app/build.gradle
gslevinkov: I would name it as "isDevNetEnabled" and that's all.
BuildConfig.DEBUG can be responsible for a lot of things, so it's better to separate flags for different things and use them.

app/build.gradle
```diff
@@ -44,6 +44,9 @@ android {
             buildConfigField("String", "intercomAppId", "\"\"")
             buildConfigField("String", "amplitudeKey", "\"\"")
         }
+        buildConfigField("boolean", "CRASHLYTICS_ENABLED", "false")
```
1sakov, line 47: And btw, maybe we should leave these fields in build types? Since we are going switch on/off crashlytics for feature builds. 

And maybe we'll do some actions with analytics

kabramovp2p: I think we should keep default items in default config
and switch them to not default on needed build type

kabramovp2p: but we can duplicate it in every buildType :) as we decide I think ))

kabramovp2p: yes yes.. but we need to add this field in all buildTypes

so we can make 3 points where we will assign fields

for every buildType

or make it default and enable or disable in needed

so in this current case according to Firebase we can make it enabled by default and disable it on debug buildType :)

or we will need to specify it in every buildType

gslevinkov: im okay to put buildConfig values in the defaultConfig with default values and override them in the buildTypes if needed

app/src/main/java/org/p2p/wallet/common/AppSettings.kt
```diff
@@ -0,0 +1,25 @@
...
+        get() = sharedPreferences.getBoolean(KEY_POLLING_ENABLED, true)
+
+    fun setPollingEnabled(isEnabled: Boolean) {
+        sharedPreferences.edit { putBoolean(KEY_POLLING_ENABLED, isEnabled) }
+    }
+
+    val isProd: Boolean
```
gslevinkov, line 19: isProduction / isReleaseBuild

app/src/main/java/org/p2p/wallet/common/AppSettings.kt
```diff
@@ -0,0 +1,25 @@
+package org.p2p.wallet.common
+
+import android.content.SharedPreferences
+import androidx.core.content.edit
+import org.p2p.wallet.BuildConfig
+
+private const val KEY_POLLING_ENABLED = "KEY_POLLING_ENABLED"
+private const val KEY_IS_PROD = "KEY_IS_PROD"
+
+class AppSettings(private val sharedPreferences: SharedPreferences) {
```
gslevinkov, line 10: It's more like AppFeatureFlags / AppFlags, but not the settings I think

app/src/main/res/layout/view_debug_drawer_configure_environment.xml
```diff
@@ -39,4 +39,15 @@
         android:text="Enable auto update"
         tools:ignore="HardcodedText" />
 
+    <androidx.appcompat.widget.SwitchCompat
```
gslevinkov, line 42: Why `Switch` is not okay here?

app/build.gradle
kabramovp2p: less lines in build config 😉

app/src/main/res/layout/view_debug_drawer_configure_environment.xml
kabramovp2p: I think it is ok... it was just copy paste here... but I agree we need to get rid from this not needed androidX prefixes

app/src/main/java/org/p2p/wallet/common/AppSettings.kt
kabramovp2p: isDevNetEnabled 😉

app/src/main/java/org/p2p/wallet/common/AppSettings.kt
kabramovp2p: it depends of usage
but yes for now I can rename it

app/src/main/java/org/p2p/wallet/common/AppSettings.kt
gslevinkov: nice!

1sakov: Mb isDevnetEnabled? 
Otherwise it looks sOmetHinGLiKe tHaT )

Bezjamper: as you wish

kabramovp2p: as you wish

kabramovp2p: yes i will
i will ping you
making changes right now

kabramovp2p: @1sakov you can check gradle please

app/build.gradle
```diff
@@ -44,6 +44,9 @@ android {
             buildConfigField("String", "intercomAppId", "\"\"")
             buildConfigField("String", "amplitudeKey", "\"\"")
         }
+        buildConfigField("boolean", "CRASHLYTICS_ENABLED", "true")
```
1sakov, line 47: Sorry but we don't need crashlytics on for debug builds, only for feature and production

1sakov (approved):
@kabramovp2p Minor comments and feel free to merge it after

PR #389, opened 2022-03-25
PWN-3109 PWN-3110 - TransactionHistory list bugfixes

gslevinkov:
## Jira Ticket (Mandatory)

https://p2pvalidator.atlassian.net/browse/PWN-3109
https://p2pvalidator.atlassian.net/browse/PWN-3110

- fixed jumping in the middle when opening the history transaction list
- fixed cropping the last element of the list
it's better to look through the separate commits


app/src/main/java/org/p2p/wallet/history/ui/history/HistoryFragment.kt
```diff
@@ -36,88 +39,101 @@ import org.p2p.wallet.utils.viewbinding.viewBinding
...
         )
     }
 
     private val binding: FragmentHistoryBinding by viewBinding()
+
     private val receiveAnalytics: ReceiveAnalytics by inject()
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
-        with(binding) {
```
gslevinkov, line 66: nothing special here, just splitted all this logic into the methods

app/src/main/java/org/p2p/wallet/history/ui/history/adapter/HistoryAdapter.kt
```diff
@@ -1,123 +1,116 @@
...
     }
 
-    override fun getItemViewType(position: Int): Int = when {
-        !stateRequiresExtraItem(pagingState) || position < itemCount - 1 -> getAdapterViewType(position)
-        pagingState is PagingState.Loading || pagingState is PagingState.InitialLoading -> R.layout.item_progress
-        else -> R.layout.item_error
+    override fun getItemViewType(position: Int): Int {
+        val preLastItemPosition = itemCount - 1
+
+        // todo: refactor to AdapterDelegate
```
gslevinkov, line 83: can I create a tech-task for this?

app/src/main/java/org/p2p/wallet/history/ui/history/adapter/holders/ProgressViewHolder.kt
```diff
@@ -0,0 +1,10 @@
+package org.p2p.wallet.history.ui.history.adapter.holders
+
+import android.view.ViewGroup
+import org.p2p.wallet.databinding.ItemProgressBinding
+import org.p2p.wallet.utils.viewbinding.inflateViewBinding
+
+class ProgressViewHolder(
```
gslevinkov, line 7: just moved all these holders to a package `holders`, it's a shame that git thinks it's a new file

app/src/main/java/org/p2p/wallet/history/ui/history/adapter/HistoryAdapterPagingController.kt
```diff
@@ -0,0 +1,35 @@
...
+            hasExtraItem && !shouldHasExtraItem -> adapter.notifyItemRemoved(count - 1)
+            !hasExtraItem && shouldHasExtraItem -> adapter.notifyItemInserted(count)
+        }
+    }
+
+    fun stateRequiresLoadingItem(state: PagingState = currentPagingState): Boolean =
+        state !in noAdditionalItemRequiredState
+
+    fun isPagingInProgress(): Boolean =
+        currentPagingState is PagingState.Loading || currentPagingState is PagingState.InitialLoading
```
1sakov, line 34: We don't need `ProgressViewHolder` if state is `InitialLoading` 
We need progress element only if user loads second page or more

app/src/main/java/org/p2p/wallet/history/ui/history/adapter/HistoryAdapterPagingController.kt
```diff
@@ -0,0 +1,35 @@
+package org.p2p.wallet.history.ui.history.adapter
+
+import org.p2p.wallet.common.ui.recycler.PagingState
+
+class HistoryAdapterPagingController(
```
1sakov, line 5: Are you sure this worth it? 
Maybe it could be useful functions in these be called in all paging adapters but if it's only within HistoryAdapter, then this logic might be more helpful to understand the adapter's logic

app/src/main/java/org/p2p/wallet/history/ui/history/adapter/HistoryAdapter.kt
```diff
@@ -1,123 +1,125 @@
...
+        //  and create item list with progress item inside instead of implicitly adding it here
+        return when {
+            !pagingController.stateRequiresLoadingItem() || position < preLastItemPosition -> getAdapterViewType(
+                position
+            )
```
1sakov, line 88: If it exceeds max line, mb it'd be better to move the whole function to new line? 
I remeber we don't wrap lines like that

app/src/main/java/org/p2p/wallet/history/ui/history/adapter/HistoryAdapter.kt
1sakov: Why do you want to move the implementation to separate class? 
Isn't simple enough to be here?

app/src/main/java/org/p2p/wallet/utils/viewbinding/ViewBindingExtenstions.kt
```diff
@@ -0,0 +1,205 @@
+package org.p2p.wallet.utils.viewbinding
```
1sakov, line 1: `ViewBindingExtenstions` -> `ViewBindingExtensions`

app/src/main/res/layout/fragment_history.xml
```diff
@@ -116,6 +116,8 @@
                 android:clipToPadding="false"
                 android:orientation="vertical"
                 android:paddingTop="24dp"
+                android:paddingBottom="@dimen/history_transaction_list_last_item_padding"
+                app:layoutManager="androidx.recyclerview.widget.LinearLayoutManager"
```
1sakov, line 120: I would add layout manager in code, it would be more obvious for the developer to understand how is recycler looked like

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryFragment.kt
```diff
@@ -36,88 +39,101 @@ import org.p2p.wallet.utils.viewbinding.viewBinding
...
 
-    private val token: Token.Active by args(EXTRA_TOKEN)
+    private val tokenForHistory: Token.Active by args(EXTRA_TOKEN)
 
     private val historyAdapter: HistoryAdapter by unsafeLazy {
         HistoryAdapter(
-            onTransactionClicked = { presenter.onItemClicked(it) },
-            onRetryClicked = { presenter.fetchNextPage() }
+            onTransactionClicked = presenter::onItemClicked,
+            onRetryClicked = presenter::fetchNextPage
```
1sakov, line 61: If you look into the bytecode, these type of calls use reflection. 
It's okay to use it in some cases but not in the whole app

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/history/ui/history/adapter/HistoryAdapter.kt
gslevinkov: It's a common practice to use AdapterDelegate pattern if you want to support different views in one list
and yes, it's not simple if we are doing it this way :)
check Epoxy or Groupie

app/src/main/java/org/p2p/wallet/history/ui/history/adapter/HistoryAdapterPagingController.kt
gslevinkov: so what should I do? it's not my check btw

Screen Shot 2022-03-25 at 9 28 42 pm

app/src/main/java/org/p2p/wallet/history/ui/history/adapter/HistoryAdapterPagingController.kt
gslevinkov: Yea, im sure
smth should control and hold logic of PagicState of this adapter.
adapter has enough responsibility and logic inside already (creating, binding, adding different type of views), so it's okay to decompose it

app/src/main/res/layout/fragment_history.xml
gslevinkov: I would rather add it here, 
it's always a boilerplate that even 3rd party libraries hide inside their own RecyclerViews. If you can declare smth in xml you should do it and remove these lines from the Fragment that has a lot already

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryFragment.kt
gslevinkov: hmm, really? i've never known. i'll fix it ofc

gslevinkov: but no reflection found here...
what can you say in your defence??


Screen Shot 2022-03-25 at 9 38 48 pm

kabramovp2p (approved):
fix comments from @1sakov 
LGTM from my side

app/src/main/java/org/p2p/wallet/history/ui/history/adapter/HistoryAdapter.kt
1sakov: Let's discuss the delegate's necessity on android meeting

app/src/main/java/org/p2p/wallet/history/ui/history/adapter/HistoryAdapterPagingController.kt
1sakov: Yes, could you remove InitialLoading state here and add this element to `noAdditionalItemRequiredState` please

app/src/main/res/layout/fragment_history.xml
1sakov: You see all params for recycler including layout manager in Fragment and remove all lines again if necessary as you said.
From the other side, setting layout manager in xml is native sdk solution and that's ok to use it.
In general, it's not a critical question, I gave you a case to think about. Please leave it if you don't think to move it is neccesary

gslevinkov: What do our colleagues have to say? @kabramovp2p @kuznetsov @davran1001

kabramovp2p: i do not like setting LinearLayoutManager in xml
so i am against of this =)

davran1001: it depends on situation i usually prefer to use xml for this

davran1001 (approved):

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryFragment.kt
1sakov: Couldn't find on my side. Let's merge it. Will come back after workaround

1sakov (approved):

PR #393, opened 2022-03-28
NO_TICKET - Crashlytics feature build fixes

gslevinkov:
## Jira Ticket (Mandatory)
no_ticket

No crash events or no-fatal errors were sent to Crashlytics when `minifyEnabled` was turned on. Strange, I tried to fix this behaviour.


kabramovp2p:
NO_TICKET
i think we should create tickets

kabramovp2p (approved):

PR #396, opened 2022-03-29
PWN-3065 - update main screen after environment is changed

gslevinkov:
## Jira Ticket (Mandatory)

https://p2pvalidator.atlassian.net/browse/PWN-305

I tried to split the new "ViewState" changes and some renamings but haven't managed to do it properly. Please forgive me.
So, the main thing here is that now we have presenter's ViewState that caches the main things that Fragment needs. 
- When we launch Fragment the first time - the presenter loads tokens, update local storage and starts polling for tokens. Also, it subscribes to env changes.
- When we leave Fragment (onDestoyView), polling stops, but the presenter still lives with cached values. Env changes subscription lives on.
- When we change env, the presenter calls the `refresh` method and updates ONLY local storage with the new env.
- When we select Fragment next time (onCreateView) - it starts polling for tokens and gets new tokens from the database.

Feel free to curse and write angry comments :)

### DIFF
--- a/app/src/main/java/org/p2p/wallet/home/ui/main/HomeFragment.kt
+++ b/app/src/main/java/org/p2p/wallet/home/ui/main/HomeFragment.kt
@@ -19,7 +19,6 @@ import org.p2p.wallet.home.analytics.BrowseAnalytics
 import org.p2p.wallet.home.model.HomeElementItem
 import org.p2p.wallet.home.model.Token
 import org.p2p.wallet.home.model.VisibilityState
-import org.p2p.wallet.home.ui.main.adapter.OnHomeItemsClickListener
 import org.p2p.wallet.home.ui.main.adapter.TokenAdapter
 import org.p2p.wallet.home.ui.select.bottomsheet.SelectTokenBottomSheet
 import org.p2p.wallet.intercom.IntercomService
@@ -30,6 +29,7 @@ import org.p2p.wallet.swap.ui.orca.OrcaSwapFragment
 import org.p2p.wallet.utils.SpanUtils
 import org.p2p.wallet.utils.getColor
 import org.p2p.wallet.utils.replaceFragment
+import org.p2p.wallet.utils.unsafeLazy
 import org.p2p.wallet.utils.viewbinding.viewBinding
 import java.math.BigDecimal
 import kotlin.math.absoluteValue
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/p2p/wallet/home/ui/main/HomeFragment.kt
```diff
@@ -39,82 +39,95 @@ private const val KEY_REQUEST_TOKEN = "KEY_REQUEST_TOKEN"
...
+    private val mainAdapter: TokenAdapter by unsafeLazy {
         TokenAdapter(this)
     }
 
     private val browseAnalytics: BrowseAnalytics by inject()
+
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
-        with(binding) {
```
gslevinkov, line 61: oh i forgot that i automatically extracted setupView method :)
sry..?

app/src/main/java/org/p2p/wallet/home/ui/main/HomePresenter.kt
```diff
@@ -219,43 +244,31 @@ class HomePresenter(
...
+        }
+    }
+
+    private fun getUserBalance(): BigDecimal =
+        presenterState.tokens
             .mapNotNull { it.totalInUsd }
             .fold(BigDecimal.ZERO, BigDecimal::add)
             .scaleShort()
 
-    private fun mapTokens(
```
gslevinkov, line 228: extracted to the mapper

app/src/main/java/org/p2p/wallet/infrastructure/network/environment/EnvironmentManager.kt
```diff
@@ -16,7 +17,11 @@ class EnvironmentManager(
     private val sharedPreferences: SharedPreferences
 ) {
 
-    private var onChanged: ((Environment) -> Unit)? = null
+    fun interface EnvironmentManagerListener {
+        fun onEnvironmentChanged(newEnvironment: Environment)
+    }
+
+    private var listeners = mutableMapOf<String, EnvironmentManagerListener>()
```
gslevinkov, line 24: can be val, gonna change it

gslevinkov:
gonna fix the unit tests

davran1001 (approved):

app/src/main/java/org/p2p/wallet/home/ui/main/HomeFragment.kt
```diff
@@ -39,82 +39,95 @@ private const val KEY_REQUEST_TOKEN = "KEY_REQUEST_TOKEN"
...
-                    replaceFragment(ReceiveSolanaFragment.create(null))
-                }
-                onSendClickListener = {
-                    replaceFragment(SendFragment.create())
-                }
-                onSwapItemClickListener = {
-                    replaceFragment(OrcaSwapFragment.create())
-                }
-            }
+        binding.setupView()
```
kabramovp2p, line 61: suggest to use name binding.bindView()

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/home/ui/main/HomeFragment.kt
gslevinkov: it already binded by viewBinding() :) 
I can rename to initView

PR #413, opened 2022-04-07
Update pull_request_template.md

kuznetsov:
#### Description

Remove screenshots table, turn Jira ticket to hyperlink, reduced titles text size

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,13 +1,7 @@
-## Jira Ticket (Mandatory)
+#### Ticket
 
-https://p2pvalidator.atlassian.net/browse/
+[PWN-XXXX](https://p2pvalidator.atlassian.net/browse/XXXX)
 
-## Description of Work (Mandatory)
+#### Description
 
-[ what work was done in this ticket ]
-
-## Screenshots (Optional)
-
-Screenshot #1 | Screenshot #2 | Screenshot #3
-:------------:|:-------------:|:------------:
-[Paste Pic 1] | [Paste Pic 2] | [Paste Pic 3]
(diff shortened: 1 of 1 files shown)


kabramovp2p (approved):

kabramovp2p:
Nice 👍

1sakov:
Why are you removing placeholders for screenshots?

kuznetsov:
> Why are you removing placeholders for screenshots?

Looks like they're used less frequently then don't, so we often should remove template text manually

.github/pull_request_template.md
```diff
@@ -1,13 +1,7 @@
-## Jira Ticket (Mandatory)
+#### Ticket
 
-https://p2pvalidator.atlassian.net/browse/
+[PWN-XXXX](https://p2pvalidator.atlassian.net/browse/XXXX)
```
gslevinkov, line 3: I liked the full hyperlink more

kuznetsov: Why? 😶

1sakov:
> > Why are you removing placeholders for screenshots?
> 
> Looks like they're used less frequently then don't, so we often should remove template text manually

If there no screenshots, you don't need to remove placeholders. 


In general we should practice more using screenshots and we'll need to create placeholders for screenshots manually then, that's more work

.github/pull_request_template.md
1sakov: The ticket number is redundant here. We use it in the title. 
In the description we just need to switch to Jira quickly, therefore there's only one action from develop: paste the ticket number which he used in the PR title

PR #416, opened 2022-04-11
PWN-1578 - Transaction History main feature

davran1001:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-1578

## Description of Work

Transaction History main feature


kuznetsov (approved):

kabramovp2p (approved):

gslevinkov:
_Co-authored-by: “khasanov.davran” <“[khasanov.davran@p2p.org](mailto:khasanov.davran@p2p.org)”>_?

kabramovp2p:
> _Co-authored-by: “khasanov.davran” <“[khasanov.davran@p2p.org](mailto:khasanov.davran@p2p.org)”>_?

yes lets get rid of this authored copy rights =)

app/src/main/java/org/p2p/wallet/history/ui/token/TokenHistoryFragment.kt
```diff
@@ -0,0 +1,178 @@
...
+        setNavigationOnClickListener { popBackStack() }
+        if (BuildConfig.DEBUG) {
+            inflateMenu(R.menu.menu_history)
+            setOnMenuItemClickListener {
+                var isHandled = false
+                if (it.itemId == R.id.closeItem) {
+                    presenter.closeAccount()
+                    isHandled = true
+                }
+                isHandled
```
1sakov, line 107: wtf? 
just return boolean directly

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/history/ui/token/TokenHistoryFragment.kt
gslevinkov: well that's my code. I couldn't understand when and what should this lambda return. What does it mean to return false or return true. So I created this val to make it explicit. Can be inlined, if there is need.

1sakov: But isn't it understandable when you return it directly?
```
{
      if (someBoolean) {
            // do smth
            return@someListener true
      }
      
      return@someListener false
}
```

the boolean variable is redundant here

PR #424, opened 2022-04-15
PWN-3114 - Device token send retry

kuznetsov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3114

## Description of Work

Retry device token send request after a minute in case of ServerException

### DIFF
--- a/app/src/main/java/org/p2p/wallet/push_notifications/ineractor/PushNotificationsInteractor.kt
+++ b/app/src/main/java/org/p2p/wallet/push_notifications/ineractor/PushNotificationsInteractor.kt
@@ -14,7 +14,7 @@ import timber.log.Timber
 
 private const val KEY_DEVICE_TOKEN = "KEY_DEVICE_TOKEN"
 private const val TOKEN_SEND_RETRY_DELAY_MS = 60000L
-private const val RETRIES_NUMBER = 1
+private const val DEFAULT_RETRIES_NUMBER = 1
 
 class PushNotificationsInteractor(
     private val deviceTokenRepository: DeviceTokenRepository,
@@ -23,7 +23,7 @@ class PushNotificationsInteractor(
     private val sharedPreferences: SharedPreferences
 ) {
 
-    suspend fun updateDeviceToken(retries: Int = RETRIES_NUMBER) {
+    suspend fun updateDeviceToken(retries: Int = DEFAULT_RETRIES_NUMBER) {
         if (retries < 0) return
 
         val token = pushTokenRepository.getPushToken().value
--- a/app/src/main/java/org/p2p/wallet/push_notifications/ineractor/PushNotificationsInteractor.kt
+++ b/app/src/main/java/org/p2p/wallet/push_notifications/ineractor/PushNotificationsInteractor.kt
@@ -3,6 +3,10 @@ package org.p2p.wallet.push_notifications.ineractor
 import android.content.SharedPreferences
 import android.os.Build
 import androidx.core.content.edit
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import org.p2p.wallet.common.di.AppScope
+import org.p2p.wallet.infrastructure.network.data.ServerException
 import org.p2p.wallet.infrastructure.network.provider.TokenKeyProvider
 import org.p2p.wallet.push_notifications.model.DeviceInfo
 import org.p2p.wallet.push_notifications.model.DeviceToken
@@ -10,15 +14,17 @@ import org.p2p.wallet.push_notifications.repository.DeviceTokenRepository
 import org.p2p.wallet.push_notifications.repository.PushTokenRepository
 
 private const val KEY_DEVICE_TOKEN = "KEY_DEVICE_TOKEN"
+private const val TOKEN_SEND_RETRY_DELAY = 60000L // ms
 
 class PushNotificationsInteractor(
     private val deviceTokenRepository: DeviceTokenRepository,
     private val pushTokenRepository: PushTokenRepository,
     private val tokenKeyProvider: TokenKeyProvider,
     private val sharedPreferences: SharedPreferences,
+    private val appScope: AppScope
 ) {
 
-    suspend fun updateDeviceToken(): DeviceToken {
+    suspend fun updateDeviceToken(retries: Int = 1) {
         val token = pushTokenRepository.getPushToken().value
 
         sharedPreferences.edit { putString(KEY_DEVICE_TOKEN, token) }
@@ -35,7 +41,20 @@ class PushNotificationsInteractor(
             clientId = tokenKeyProvider.publicKey
         )
 
-        return deviceTokenRepository.sendDeviceToken(deviceToken)
+        try {
+            deviceTokenRepository.sendDeviceToken(deviceToken)
+        } catch (e: ServerException) {
+            retryTokenSend(retries)
+        }
+    }
+
+    private fun retryTokenSend(retries: Int) {
+        if (retries < 1) return
+
+        appScope.launch {
+            delay(TOKEN_SEND_RETRY_DELAY)
+            updateDeviceToken(retries - 1)
+        }
     }
 
     suspend fun deleteDeviceToken() {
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/p2p/wallet/push_notifications/ineractor/PushNotificationsInteractor.kt
```diff
@@ -3,22 +3,28 @@ package org.p2p.wallet.push_notifications.ineractor
...
 class PushNotificationsInteractor(
     private val deviceTokenRepository: DeviceTokenRepository,
     private val pushTokenRepository: PushTokenRepository,
     private val tokenKeyProvider: TokenKeyProvider,
     private val sharedPreferences: SharedPreferences,
+    private val appScope: AppScope
 ) {
 
-    suspend fun updateDeviceToken(): DeviceToken {
+    suspend fun updateDeviceToken(retries: Int = 1) {
```
kabramovp2p, line 27: i'd prefer to move 1 to constant and use/reuse it

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/push_notifications/ineractor/PushNotificationsInteractor.kt
kuznetsov: It's a one use value, so it's no need for constant

davran1001: Agree with @kabramovp2p, move to variable or to constant

app/src/main/java/org/p2p/wallet/push_notifications/ineractor/PushNotificationsInteractor.kt
```diff
@@ -35,7 +41,20 @@ class PushNotificationsInteractor(
...
+            retryTokenSend(retries)
+        }
+    }
+
+    private fun retryTokenSend(retries: Int) {
+        if (retries < 1) return
+
+        appScope.launch {
+            delay(TOKEN_SEND_RETRY_DELAY)
+            updateDeviceToken(retries - 1)
```
davran1001, line 56: maybe we move retries - 1 to separate line ?

app/src/main/java/org/p2p/wallet/push_notifications/ineractor/PushNotificationsInteractor.kt
```diff
@@ -3,22 +3,28 @@ package org.p2p.wallet.push_notifications.ineractor
...
+import org.p2p.wallet.common.di.AppScope
+import org.p2p.wallet.infrastructure.network.data.ServerException
 import org.p2p.wallet.infrastructure.network.provider.TokenKeyProvider
 import org.p2p.wallet.push_notifications.model.DeviceInfo
 import org.p2p.wallet.push_notifications.model.DeviceToken
 import org.p2p.wallet.push_notifications.repository.DeviceTokenRepository
 import org.p2p.wallet.push_notifications.repository.PushTokenRepository
 
 private const val KEY_DEVICE_TOKEN = "KEY_DEVICE_TOKEN"
+private const val TOKEN_SEND_RETRY_DELAY = 60000L // ms
```
gslevinkov, line 17: you can put MS in the const name or rewrite using TimeUnit

app/src/main/java/org/p2p/wallet/push_notifications/ineractor/PushNotificationsInteractor.kt
gslevinkov, line 56: you are not retrying sending token here, but rewriting shared prefs too

app/src/main/java/org/p2p/wallet/push_notifications/ineractor/PushNotificationsInteractor.kt
```diff
@@ -35,7 +41,20 @@ class PushNotificationsInteractor(
...
+    }
+
+    private fun retryTokenSend(retries: Int) {
+        if (retries < 1) return
+
+        appScope.launch {
+            delay(TOKEN_SEND_RETRY_DELAY)
+            updateDeviceToken(retries - 1)
+        }
     }
```
gslevinkov, line 58:
```suggestion
        appScope.launch {
            sendTokenToServer(deviceToken, retries)
        }
    }

    private suspend fun sendTokenToServer(deviceToken: DeviceToken, retries: Int) {
        val noRetryNeeded = retries == 0
        if (noRetryNeeded) return

        try {
            deviceTokenRepository.sendDeviceToken(deviceToken)
        } catch (e: ServerException) {
            delay(TOKEN_SEND_RETRY_DELAY)
            sendTokenToServer(deviceToken, retries - 1)
        }
    }
```

gslevinkov: your first send try was not in appScope, so I moved it
i extracted sending method so it's recursion in only one place, plus no shared preferences logic included any longer

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/push_notifications/ineractor/PushNotificationsInteractor.kt
```diff
@@ -35,7 +41,20 @@ class PushNotificationsInteractor(
             clientId = tokenKeyProvider.publicKey
         )
 
-        return deviceTokenRepository.sendDeviceToken(deviceToken)
+        try {
+            deviceTokenRepository.sendDeviceToken(deviceToken)
+        } catch (e: ServerException) {
+            retryTokenSend(retries)
```
1sakov, line 47: Please, add timber logs

app/src/main/java/org/p2p/wallet/push_notifications/ineractor/PushNotificationsInteractor.kt
kuznetsov: Okay then

app/src/main/java/org/p2p/wallet/push_notifications/ineractor/PushNotificationsInteractor.kt
kuznetsov: Thank, that'a a good point!

1sakov (approved):

app/src/main/java/org/p2p/wallet/push_notifications/ineractor/PushNotificationsInteractor.kt
```diff
@@ -3,22 +3,29 @@ package org.p2p.wallet.push_notifications.ineractor
...
 import org.p2p.wallet.infrastructure.network.provider.TokenKeyProvider
 import org.p2p.wallet.push_notifications.model.DeviceInfo
 import org.p2p.wallet.push_notifications.model.DeviceToken
 import org.p2p.wallet.push_notifications.repository.DeviceTokenRepository
 import org.p2p.wallet.push_notifications.repository.PushTokenRepository
+import timber.log.Timber
 
 private const val KEY_DEVICE_TOKEN = "KEY_DEVICE_TOKEN"
+private const val TOKEN_SEND_RETRY_DELAY_MS = 60000L
+private const val RETRIES_NUMBER = 1
```
gslevinkov, line 17:
```suggestion
private const val DEFAULT_RETRIES_NUMBER = 1
```

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/push_notifications/ineractor/PushNotificationsInteractor.kt
```diff
@@ -35,7 +42,13 @@ class PushNotificationsInteractor(
             clientId = tokenKeyProvider.publicKey
         )
 
-        return deviceTokenRepository.sendDeviceToken(deviceToken)
+        try {
+            deviceTokenRepository.sendDeviceToken(deviceToken)
+        } catch (e: ServerException) {
+            Timber.e(e, "Error sending device token $token to server")
+            delay(TOKEN_SEND_RETRY_DELAY_MS)
```
davran1001, line 49: It's not a good idea to block thread resources for one minute, i think it's insufficient way and we need to go with another way of implementation

kuznetsov: Let's not overhead the solution for now and check how it performs on the QA stage

davran1001 (approved):

PR #426, opened 2022-04-18
PWN-3251 - pull to refresh impl history

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3251

## Description of Work

pull to refresh implementation on history screen

## Screenshots
Android Emulator - Pixel_5_API_30:5554 2022-04-18 15-50-33


app/src/main/java/org/p2p/wallet/history/ui/history/HistoryFragment.kt
```diff
@@ -53,10 +62,18 @@ class HistoryFragment :
...
+        isRefreshing = state is PagingState.Loading && state.isRefresh
         adapter.setPagingState(state)
-        binding.shimmerView.isVisible = state == PagingState.InitialLoading
-        binding.refreshLayout.isRefreshing = state is PagingState.Loading && state.isRefresh
-        binding.refreshLayout.isVisible = state != PagingState.InitialLoading
+        with(binding) {
+            shimmerView.isVisible = state == PagingState.InitialLoading
+            refreshLayout.isRefreshing = state is PagingState.Loading && state.isRefresh
+            refreshLayout.isVisible = state != PagingState.InitialLoading
+            historyRecyclerView.alpha = if (isRefreshing) {
```
1sakov, line 71: I think it's not the best way. 
Would be better to add a fullscreen view with half-transparent background and make it clickable to prevent clicks. 
And just show it when it's refreshing. 

Otherwise it's only simulating instead of real behavior. 
It maybe broken if some accidentally delete checking for `isRefreshing` and showing the real blocking view is cheap as well

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryFragment.kt
kabramovp2p: yes, I agree, I can make it it was a second way for implementation 

ok will do it

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryFragment.kt
```diff
@@ -27,9 +30,15 @@ class HistoryFragment :
 
     override val presenter: HistoryContract.Presenter by inject()
     private val binding: FragmentHistoryBinding by viewBinding()
+
+    private var isRefreshing = false
     private val adapter: HistoryAdapter by unsafeLazy {
         HistoryAdapter(
-            onTransactionClicked = presenter::onItemClicked,
```
gslevinkov, line 32: I would suggest to remove kt-lint rule that we need to always put namedArg on a new line 
so instead of 
```kotlin
HistoryAdapter(
   onTransactionClicked = {}
)
```
we would have
```kotlin
HistoryAdapter(onTransactionClicked = {
    
})
```
no things to fix here, just a suggestion to discuss

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryFragment.kt
```diff
@@ -27,9 +30,15 @@ class HistoryFragment :
...
+            onTransactionClicked = {
+                if (!isRefreshing) {
+                    presenter.onItemClicked(it)
```
gslevinkov, line 39: we can specify for presenter that we clicked on transaction, but not on 'item' in general. Item is more of adapter layer terminology, on presenter level we know we clicked on transaction

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryFragment.kt
davran1001: we can create something like Fullscreen progress view

kabramovp2p: ?

kabramovp2p:
@1sakov Updated

davran1001 (approved):

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryFragment.kt
gslevinkov: didn't understand why it's resolved, but ok...?

gslevinkov (approved):

1sakov (approved):

PR #436, opened 2022-04-22
PWN-3355 - Real NotificationService integration

kuznetsov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3355

## Description of Work

- Replace stub NotificationService endpoint with real one
- Remove `RpcInterceptor` from NotificationService Retrofit
- Put `httpLoggingInterceptor` in the end after all interceptors
- Remove useless `RpcRequestObject` class


app/src/main/AndroidManifest.xml
```diff
@@ -15,6 +15,7 @@
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.USE_BIOMETRIC" />
 
+    <!-- todo: remove usesCleartextTraffic after switching to https on NotificationService -->
```
kabramovp2p, line 18: add number of task please 
and create it not to forgot about this!

kabramovp2p (approved):

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/infrastructure/network/NetworkModule.kt
```diff
@@ -61,7 +61,7 @@ object NetworkModule : InjectionModule {
 
         single(named(NOTIFICATION_SERVICE_RETROFIT_QUALIFIER)) {
```
davran1001, line 62: I am not sure that we need separate Retrofit instance for NotificationService, cause we don't use base url in requests

app/src/main/AndroidManifest.xml
1sakov: We already have it https://p2pvalidator.atlassian.net/browse/PWN-3208

app/src/main/java/org/p2p/wallet/push_notifications/api/NotificationServiceApi.kt
```diff
@@ -1,24 +1,20 @@
...
-        @Body rpcRequest: RpcRequestObject,
-        @Url url: String = emptyString()
+        @Body rpcRequest: RpcRequest
     ): CommonResponse<DeviceToken>
 
-    @POST
+    @POST("notifier")
     suspend fun deleteDeviceToken(
-        @Body rpcRequest: RpcRequestObject,
-        @Url url: String = emptyString()
```
davran1001, line 22: You can use this parameter to avoid creating new retrofit instance

app/src/main/java/org/p2p/wallet/push_notifications/model/DeviceToken.kt
```diff
@@ -3,6 +3,8 @@ package org.p2p.wallet.push_notifications.model
 import com.google.gson.annotations.SerializedName
 
 data class DeviceToken(
```
davran1001, line 5: By the name it's not obvious to determine, is it Request body or it's response body, cause if it's response body i don't recommend to use default values

app/src/main/java/org/p2p/wallet/push_notifications/api/NotificationServiceApi.kt
```diff
@@ -1,24 +1,20 @@
...
 import org.p2p.wallet.push_notifications.model.DeviceToken
-import org.p2p.wallet.utils.emptyString
 import retrofit2.http.Body
 import retrofit2.http.POST
-import retrofit2.http.Url
 
 interface NotificationServiceApi {
 
-    @POST
+    @POST("notifier")
```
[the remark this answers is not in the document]
gslevinkov: it's not dynamic, so I don't think @Url is a good place

PR #441, opened 2022-04-25
PWN-3401 - Add error state for history screen

davran1001:
## Jira Ticket 

https://p2pvalidator.atlassian.net/browse/PWN-3401

## Description of Work 

Add error state view for history screen

### DIFF
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -47,8 +47,6 @@
     <color name="backgroundBannerSecondary">#D7D1FF</color>
     <color name="colorBackgroundNotification">#404040</color>
 
-    <color name="secondaryGray">#6F7D8D</color>
-
     <!--Android Overridden colors-->
     <color name="mtrl_textinput_default_box_stroke_color" tools:override="true">@color/backgroundDisabled</color>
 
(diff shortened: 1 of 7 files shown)


kabramovp2p (approved):

app/src/main/res/values/colors.xml
```diff
@@ -47,6 +47,8 @@
     <color name="backgroundBannerSecondary">#D7D1FF</color>
     <color name="colorBackgroundNotification">#404040</color>
 
+    <color name="secondaryGray">#6F7D8D</color>
```
1sakov, line 50: I am not sure if we have such color in our design system

1sakov:
@davran1001 minor comments. Please use the existing colors

1sakov (approved):

app/src/main/res/values/colors.xml
kabramovp2p: i think new feature is not correctly added according our design system
we need to aks Gena to add all colors to design system or change designs according to our design system =(

1sakov: The current history design doesn't fit the general design. 
Secondary gray color is `textIconSecondary`

Please work with designers and ask to change it. 
The history design does have wrong font as well

gslevinkov (approved):

PR #442, opened 2022-04-25
PWN-3220 - Fix progress bottom sheet state

davran1001:
## Jira Ticket 

https://p2pvalidator.atlassian.net/browse/PWN-3220

## Description of Work 

Fix progress bottom sheet state when user go out from screen and go back

### DIFF
--- a/app/src/main/java/org/p2p/wallet/transaction/TransactionManager.kt
+++ b/app/src/main/java/org/p2p/wallet/transaction/TransactionManager.kt
@@ -38,7 +38,7 @@ class TransactionManager private constructor(
     /*
    * This one is to show user transaction id when progress dialog is shown
    * */
-    private val transactionStateFlow = MutableStateFlow<TransactionState?>(null)
+    private val transactionStateFlow = MutableStateFlow<TransactionState>(TransactionState.Progress())
 
     private val pendingTransactions = mutableListOf<AppTransaction>()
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/p2p/wallet/transaction/TransactionManager.kt
```diff
@@ -37,7 +38,7 @@ class TransactionManager private constructor(
     /*
    * This one is to show user transaction id when progress dialog is shown
    * */
-    private val transactionStateFlow = MutableStateFlow<TransactionState>(TransactionState.Progress())
+    private val transactionStateFlow = MutableStateFlow<TransactionState?>(null)
```
gslevinkov, line 41: are we sure we don't want to create a separate state instead of NULL?
Explicit Separate class is always better then NULL

davran1001: I even don't know how name this state x D  None ?

gslevinkov: None / Initial / Empty / Nothing

gslevinkov: any synomym that can used instead of NULL. NULL - means absence of something

davran1001: i used null cause null value will not trigger Flow again

davran1001: it's behavior which i wanted to reach to fix this problem x D

gslevinkov: you can try and filter new separate class like null

davran1001:
i used wrong ticket number for branch name

1sakov (approved):

app/src/main/java/org/p2p/wallet/transaction/TransactionManager.kt
```diff
@@ -68,9 +68,12 @@ class TransactionManager private constructor(
         return executor.getStateFlow()
     }
 
-    fun getTransactionStateFlow(): Flow<TransactionState> = transactionStateFlow
-
-    fun getTransactionState(): TransactionState = transactionStateFlow.value
+    fun getTransactionStateFlow(refresh: Boolean = true): Flow<TransactionState> {
```
gslevinkov, line 71: refresh  -> forceIsProgress 
because you are force changing the value inside flow

gslevinkov (approved):

kuznetsov (approved):

PR #455, opened 2022-04-27
PWN-3447 Disable sending SPL tokens to renBTC and vice versa

kuznetsov:
### Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3447

### Description of Work

Show error if BTC address was pasted in Solana network and vice versa

### Screenshots

Screenshot 2022-04-27 at 19 10 54 | Screenshot 2022-04-27 at 19 17 05
:------------:|:-------------:


app/src/main/java/org/p2p/wallet/send/ui/main/SendPresenter.kt
```diff
@@ -152,6 +152,14 @@ class SendPresenter(
     override fun setTargetResult(result: SearchResult?) {
         target = result
 
+        result?.let {
+            selectNetworkType(result.networkType)
+            if (!token.isRenBTC && result.networkType == NetworkType.BITCOIN) {
+                view?.showWrongAddressTarget(result.address)
+                return
```
davran1001, line 159: for what return statement was used here ?

app/src/main/java/org/p2p/wallet/send/model/SearchResult.kt
```diff
@@ -4,7 +4,8 @@ import android.os.Parcelable
 import kotlinx.parcelize.Parcelize
 
 sealed class SearchResult(
-    open val address: String
+    open val address: String,
+    open val networkType: NetworkType = NetworkType.SOLANA
```
davran1001, line 8: you can remove default value here

app/src/main/java/org/p2p/wallet/send/ui/main/SendPresenter.kt
```diff
@@ -152,6 +152,14 @@ class SendPresenter(
     override fun setTargetResult(result: SearchResult?) {
         target = result
 
+        result?.let {
+            selectNetworkType(result.networkType)
+            if (!token.isRenBTC && result.networkType == NetworkType.BITCOIN) {
```
davran1001, line 157: first statement is redundant cause you check   `if (!token.isRenBTC) return` in `selectNetworkType`

app/src/main/java/org/p2p/wallet/send/ui/main/SendPresenter.kt
```diff
@@ -152,6 +152,14 @@ class SendPresenter(
     override fun setTargetResult(result: SearchResult?) {
```
davran1001, line 152: you can remove ? from this method, and just avoid calling this method if result is null, cause inside this function you anyway checking for nullable

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/send/model/SearchResult.kt
kuznetsov: I've got your point. But the thing is that networkType is changed in AddressOnly results in Bitcoin case, so to not set NetworkType.SOLANA in most places I just used default value

app/src/main/java/org/p2p/wallet/send/ui/main/SendPresenter.kt
kuznetsov: It can't be changed cause in SearchFragment we call this functions with null param
```
            clearImageView.setOnClickListener {
                presenter.setTargetResult(null)
            }
```

app/src/main/java/org/p2p/wallet/send/ui/main/SendPresenter.kt
davran1001: so you have not another statements

app/src/main/java/org/p2p/wallet/send/ui/main/SendPresenter.kt
```diff
@@ -695,4 +715,7 @@ class SendPresenter(
 
     private fun isAddressValid(address: String?): Boolean =
         !address.isNullOrBlank() && address.trim().length >= VALID_ADDRESS_LENGTH
+
+    private val Token.Active?.isRenBTC: Boolean
```
davran1001, line 719: Token.Active is a sealed class, so i think we should use some additional state for this class like None, or Empty to avoid using nullable

app/src/main/java/org/p2p/wallet/send/model/Target.kt
```diff
@@ -50,4 +50,8 @@ data class Target constructor(
                 else -> Validation.INVALID
             }
         }
+
+    @IgnoredOnParcel
```
davran1001, line 54: I don't see reasons to add this attribute, can you describe why you add it here ?

app/src/main/java/org/p2p/wallet/send/ui/main/SendPresenter.kt
kuznetsov: It's a good point for future refactoring

app/src/main/java/org/p2p/wallet/send/model/Target.kt
kuznetsov: I've set this annotation similarly to previous properties and consider it just as utility property with no need to be parselized and transferred

gslevinkov (approved):

1sakov (approved):

PR #458, opened 2022-04-28
PWN-3475 - History empty state

1sakov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3475

## Description of Work

Added EmptyDataException for one catch above

### DIFF
--- a/app/src/main/java/org/p2p/wallet/history/ui/history/HistoryPresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/history/ui/history/HistoryPresenter.kt
@@ -1,8 +1,5 @@
 package org.p2p.wallet.history.ui.history
 
-import kotlinx.coroutines.CancellationException
-import kotlinx.coroutines.Job
-import kotlinx.coroutines.launch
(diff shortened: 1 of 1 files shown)


gslevinkov (approved):

davran1001 (approved):

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryPresenter.kt
```diff
@@ -88,11 +88,11 @@ class HistoryPresenter(
             }
         } catch (e: CancellationException) {
             Timber.w(e, "Cancelled history next page load")
+        } catch (e: EmptyDataException) {
```
kabramovp2p, line 91: empty data exception ? :) zest'

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryPresenter.kt
1sakov: Response model complexity.
But we can workaround.

PR #463, opened 2022-05-03
PWN-3511 - Libs version update

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3511

## Description of Work

PWN-3511 - kotlinVersion - 1.6.10 -> 1.6.21, build:gradle - 7.1.2 -> 7.1.3, roomVersion - 2.4.1 -> 2.4.2, org.robolectric:robolectric - 4.5.1 -> 4.6.1

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -217,7 +217,7 @@ dependencies {
     implementation 'com.caverock:androidsvg-aar:1.4'
 
     // Room
-    def roomVersion = '2.4.1'
+    def roomVersion = '2.4.2'
     implementation "androidx.room:room-runtime:$roomVersion"
     implementation "androidx.room:room-ktx:$roomVersion"
     kapt "androidx.room:room-compiler:$roomVersion"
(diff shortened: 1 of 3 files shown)


app/build.gradle
```diff
@@ -242,7 +242,7 @@ dependencies {
     // https://mvnrepository.com/artifact/junit/junit
     testImplementation 'junit:junit:4.13.2'
     // https://github.com/robolectric/robolectric
-    testImplementation "org.robolectric:robolectric:4.5.1"
+    testImplementation "org.robolectric:robolectric:4.6.1"
```
1sakov, line 245: Don't do it, tests will fail. 
It's bugged version, all devs are using 4.5.1

1sakov (changes requested):

app/build.gradle
kabramovp2p: ohh I see 
ok will revert it 👌🏻👍🏻

gslevinkov: why do we need robolectric:robolectric at first place?

gslevinkov (approved):

app/build.gradle
gslevinkov: I meant do we use roboelectic in our current unit tests?

kabramovp2p: @1sakov ^^ ?

kabramovp2p: @1sakov fixed! ^^

1sakov: we do

1sakov (approved):

kuznetsov (approved):

davran1001 (approved):

PR #466, opened 2022-05-04
PWN-3266 - Show user card in gallery

kuznetsov:
### Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3266

### Description of Work

- Save user card image to private `Pictures` directory
- Add saved image to gallery

### Screenshots

Screenshot 2022-05-04 at 17 01 55


app/src/main/java/org/p2p/wallet/auth/repository/FileRepository.kt
```diff
@@ -44,7 +45,7 @@ class FileRepository(private val context: Context) {
         val fileName = name ?: Date().toString()
         try {
             val mainDir = File(
-                context.getExternalFilesDir(Environment.DIRECTORY_PICTURES),
+                Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_PICTURES),
```
gslevinkov, line 48: let's do Dependency Injection here, shall we? would be much easier later on

1sakov: Are you sure it's working on all OS since this method is deprecated?

kuznetsov: It works on latest OS

kuznetsov: We will test on earlier version on QA

gslevinkov (approved):

davran1001 (approved):

PR #467, opened 2022-05-05
NO-TICKET - Debug view visibility persistence

1sakov:


kabramovp2p (approved):
yuhoo! thx

app/src/main/java/org/p2p/wallet/root/RootActivity.kt
```diff
@@ -105,19 +110,24 @@ class RootActivity : BaseMvpActivity<RootContract.View, RootContract.Presenter>(
     @SuppressLint("SetTextI18n")
     private fun initializeDebugDrawer() {
         val devView = findViewById<TextView>(R.id.developTextView)
+        val isDebugHidden = sharedPreferences.getBoolean(KEY_DEBUG_HIDDEN, false)
```
gslevinkov, line 113:
```suggestion
        val isDebugPanelHidden = sharedPreferences.getBoolean(KEY_DEBUG_HIDDEN, false)
```

we can't hide debug itsself :)

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/root/RootActivity.kt
1sakov: but it's not panel, it's debug view

1sakov: I can rename to `isDebugViewHidden` if you don't mind @gslevinkov

davran1001 (approved):

kuznetsov (approved):

PR #470, opened 2022-05-06
PWN-3044 - fix fees on Send and Swap

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3044

## Description of Work

PWN-3044 - if isFreeTransactionAvailable no need to show user accountCreation + transaction fee use only accountCreation as main fee because feeRelayer pays for transactions


app/src/main/java/org/p2p/wallet/send/interactor/SendInteractor.kt
```diff
@@ -86,9 +86,7 @@ class SendInteractor(
                 transactionFee += lamportsPerSignature
 
                 // feePayer's signature
-                if (!feePayerToken.isSOL) {
```
kuznetsov, line 89: Why didn't you add `!isFreeTransactionAvailable()` condition here?

kuznetsov (approved):

davran1001 (approved):

davran1001 (approved):

PR #473, opened 2022-05-08
PWN-3400-History transaction parser

davran1001:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3400(In progress)

## Description of Work
Include this tasks
https://p2pvalidator.atlassian.net/browse/PWN-3457
https://p2pvalidator.atlassian.net/browse/PWN-3057
https://p2pvalidator.atlassian.net/browse/PWN-1583


app/src/main/java/org/p2p/wallet/history/interactor/mapper/SolanaInstructionParser.kt
```diff
@@ -1,125 +1,90 @@
...
-                    transactionRoot = transactionRoot
-                )
+            val parsedInfo = instruction.parsed
+            val type = parsedInfo?.type
+            when (TransactionDetailsType.valueOf(type)) {
+                TransactionDetailsType.TRANSFER, TransactionDetailsType.TRANSFER_CHECKED -> {
+                    parseTransferTransaction(
+                        signature = signature,
+                        transactionRoot = transactionRoot,
+                        parsedInfo = parsedInfo!!,
```
kabramovp2p, line 37: parsedInfo!! not safe here
we need to make it not null or check it before or make some fallback info if it will be null

kabramovp2p (approved):
LGTM but hard to check a lot of code

app/src/main/java/org/p2p/wallet/history/interactor/mapper/SolanaInstructionParser.kt
```diff
@@ -1,125 +1,90 @@
...
-                )
-            }
-            else -> {
-                parseUnknownTransaction(
-                    signature = signature,
-                    transactionRoot = transactionRoot
-                )
+            val parsedInfo = instruction.parsed
+            val type = parsedInfo?.type
+            when (TransactionDetailsType.valueOf(type)) {
```
1sakov, line 32: `type` is lowercase here but `TransactionDetailsType` are upper case. 
Please make sure that this parsing works correct

app/src/main/java/org/p2p/wallet/history/model/TransactionDetailsLaunchState.kt
```diff
@@ -5,8 +5,9 @@ import kotlinx.parcelize.Parcelize
 
 sealed class TransactionDetailsLaunchState : Parcelable {
     @Parcelize
-    data class Id(val transactionId: String) : TransactionDetailsLaunchState()
+    data class Id(val tokenPublicKey: String, val transactionId: String) : TransactionDetailsLaunchState()
```
gslevinkov, line 9: it's better to keep formatting as one for all the descendants 

```suggestion
    data class Id(val tokenPublicKey: String, val transactionId: String) : 
        TransactionDetailsLaunchState()

```

app/src/main/java/org/p2p/wallet/history/repository/remote/TransactionDetailsRpcRepository.kt
```diff
@@ -1,40 +1,82 @@
...
+                                userPublicKey = tokenPublicKey
+                            )
+                        )
                     }
                 }
+                transactionDetails.forEach {
+                    it.error = transaction.meta.error?.instructionError
+                }
+            } catch (e: Exception) {
+                Timber.e(e)
```
1sakov, line 77: Added custom note as well please

app/src/main/java/org/p2p/wallet/history/repository/remote/TransactionDetailsRpcRepository.kt
```diff
@@ -1,40 +1,82 @@
...
+        }
+    }
+
+    private fun fromNetworkToDomain(
+        tokenPublicKey: String,
+        transactions: List<ConfirmedTransactionRootResponse>
+    ): List<TransactionDetails> {
+        val transactionDetails = mutableListOf<TransactionDetails>()
+        transactions.forEach { transaction ->
+            try {
```
gslevinkov, line 49: it's better to extract this whole block to a method to remove indentation and complexity

app/src/main/java/org/p2p/wallet/history/repository/remote/TransactionDetailsRpcRepository.kt
```diff
@@ -1,40 +1,82 @@
...
+                val signature = transaction.transaction?.getTransactionId() ?: return@forEach
+                when {
+                    OrcaSwapInstructionParser.isTransactionContainsOrcaSwap(transaction) -> {
+                        val orcaSwapDetails =
+                            OrcaSwapInstructionParser.parse(signature = signature, transactionRoot = transaction)
+                        transactionDetails.add(orcaSwapDetails.getOrThrow())
+                    }
+                    SerumSwapInstructionParser.isTransactionContainsSerumSwap(transaction) -> {
```
gslevinkov, line 57: please remove static calls here and inject parser directly

app/src/main/java/org/p2p/wallet/history/repository/remote/TransactionDetailsRpcRepository.kt
gslevinkov, line 77: should we skip parsing all the other transactions if only one has failed?

app/src/main/java/org/p2p/wallet/history/ui/token/TokenHistoryFragment.kt
```diff
@@ -83,7 +83,7 @@ class TokenHistoryFragment :
         totalTextView.text = tokenForHistory.getFormattedTotal(includeSymbol = true)
         usdTotalTextView.text = tokenForHistory.getFormattedUsdTotal()
 
-        refreshLayout.setOnRefreshListener { presenter.refresh() }
+        refreshLayout.setOnRefreshListener { presenter.retry() }
```
gslevinkov, line 86: what should presenter retry? RetryLoad? retry parsing? I suggest to choose more specific and explicit name for this method

app/src/main/java/org/p2p/wallet/transaction/model/TransactionStatus.kt
```diff
@@ -10,6 +11,10 @@ enum class TransactionStatus(@StringRes val resValue: Int) {
     ERROR(R.string.details_error);
 
     companion object {
-        fun from(response: TransactionDetails) = if (response.error.isNullOrEmpty()) COMPLETED else ERROR
+        fun from(response: TransactionDetails) = when {
```
gslevinkov, line 14: explicit type

app/src/main/java/org/p2p/wallet/history/interactor/HistoryInteractor.kt
```diff
@@ -24,89 +23,41 @@ class HistoryInteractor(
...
-        forceNetwork: Boolean
-    ): List<HistoryTransaction> {
-        if (forceNetwork) {
-            return transactionsRemoteRepository.getTransactions(signatures)
-                .also { transactionsLocalRepository.saveTransactions(it) }
-                .mapToHistoryTransactions(tokenPublicKey)
-        }
+    suspend fun getHistoryTransaction(tokenPublicKey: String, transactionId: String) =
+        transactionsLocalRepository.getTransactions(listOf(transactionId)).mapToHistoryTransactions(tokenPublicKey)
+            .first()
```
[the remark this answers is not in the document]
davran1001: It's impossible, cause every time after fetching transactions we store them in DB

app/src/main/java/org/p2p/wallet/history/repository/remote/TransactionDetailsRpcRepository.kt
davran1001: For what ?, it's just objects, i don't see reasons convert them to classes

gslevinkov: It would be easier to mock their behavior later in unit tests.
instead of duck taping, we could just mock any of parsers to return true/false for their `is*` methods to check every type

app/src/main/java/org/p2p/wallet/history/repository/remote/TransactionDetailsRpcRepository.kt
davran1001: so i have already moved it to fromNetworkToDomain fun

app/src/main/java/org/p2p/wallet/history/repository/remote/TransactionDetailsRpcRepository.kt
davran1001: removed this

app/src/main/java/org/p2p/wallet/history/repository/remote/TransactionDetailsRpcRepository.kt
davran1001: for this we can just create an interface and create second mock variant of interface and use it

1sakov (approved):

PR #474, opened 2022-05-11
PWN-3408 - Fetching swap configs directly from orca api

1sakov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3408

## Description of Work
Update orca pools and tokens


kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/swap/SwapModule.kt
```diff
@@ -39,8 +38,9 @@ object SwapModule : InjectionModule {
         }
 
         single {
-            OrcaSwapInternalRemoteRepository(get(), get())
-        } bind OrcaSwapInternalRepository::class
+            val baseUrl = get<Context>().getString(R.string.orca_api_base_url)
```
gslevinkov, line 41:
```suggestion
            val baseUrl = androidContext().getString(R.string.orca_api_base_url)
```

app/src/main/java/org/p2p/wallet/swap/SwapModule.kt
```diff
@@ -39,8 +38,9 @@ object SwapModule : InjectionModule {
...
+            getRetrofit(baseUrl, "Orca", null).create(OrcaApi::class.java)
```
gslevinkov, line 42: named args please

gslevinkov (approved):

PR #475, opened 2022-05-11
WIP: PWN-3522 - apk to slack

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3522

## Description of Work

- upload APK to Firebase AND Slack

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,6 +39,20 @@ commands:
       - restore_cache:
           key: jars-{{ checksum "build.gradle" }}-{{ checksum  "app/build.gradle" }}
 
+
+  upload_apk_to_slack:
+    description: "Upload builded app to Slack channel"
+    steps:
+      - run:
+          name: "Upload using CURL request to Slack"
+          command: |
+            curl \
+              -F token="${SLACK_CHANNEL_ACCESS_TOKEN}" \
+              -F channels="U03DXGAQ4DV" \
+              -F initial_comment="Hey! Here is the APK of branch (Build number ${CIRCLE_BUILD_NUM} Success :tada:)" \
+              -F "file=@$(find app/build/outputs/apk/feature -name 'app*')" \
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -39,6 +39,34 @@ commands:
       - restore_cache:
           key: jars-{{ checksum "build.gradle" }}-{{ checksum  "app/build.gradle" }}
 
+
+  upload_feature_apk_to_slack:
```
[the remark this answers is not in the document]
1sakov: I think that's ok for now

app/build.gradle
```diff
@@ -37,6 +37,20 @@ def initApiKeys(defaultConfig) {
...
+
+    if (buildNumberFromCi != null) {
+        return VERSION_BUILD as Integer + buildNumberFromCi.toInteger()
+    } else {
+        return VERSION_BUILD
+    }
+}
+
+String getBuildAppFileName() {
+    """p2p_wallet_${getVersionBuild()}_${VERSION_NAME}_${taskNumber()}"""
```
1sakov, line 51: I don't think we need tasknumber here. 
Let's make the following format: 

"""P2P_Wallet_${VERSION_NAME} (${getVersionBuild()})"""

And can we add build type as well?

1sakov (changes requested):

1sakov:
minor comments

app/build.gradle
gslevinkov: well, that would be very painful to do because we need to create a separate GRADLE task and get build type there. 

also, about taskNumber, how should the testers understand which branch was built?

1sakov: in release notes

kabramovp2p (approved):

kuznetsov (approved):

1sakov (approved):

PR #476, opened 2022-05-11
PWN-3284 - update presenterState so banners can be updated

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3284

## Description of Work

PWN-3284 - update presenterState so banners can be updated


app/src/main/java/org/p2p/wallet/home/ui/main/HomePresenter.kt
```diff
@@ -273,7 +270,15 @@ class HomePresenter(
         return if (usernameExists) {
             listOf(feedbackBanner)
         } else {
-            listOf(usernameBanner, feedbackBanner)
+            listOf(
+                Banner(
```
gslevinkov, line 274: maybe extract this Banner creation to a local variable and then rewrite it as

listOf(mainBanner, feedbackBanner)?
i cant understand what banner this is.

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/home/ui/main/HomePresenter.kt
kabramovp2p: hmmm maybe you are right

gslevinkov (approved):

davran1001 (approved):

PR #481, opened 2022-05-13
PWN-3299 - Add username length validation

kuznetsov:
### Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3299

### Description of Work

Show error state on username longer then 15 characters

### Screenshots

Screenshot 2022-05-13 at 17 38 08

### DIFF
--- a/app/src/main/java/org/p2p/wallet/auth/ui/username/ReserveUsernamePresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/auth/ui/username/ReserveUsernamePresenter.kt
@@ -31,7 +31,6 @@ class ReserveUsernamePresenter(
             return
         }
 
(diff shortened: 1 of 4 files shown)


kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/auth/ui/username/ReserveUsernamePresenter.kt
```diff
@@ -29,6 +31,12 @@ class ReserveUsernamePresenter(
             return
         }
 
+        // check max length
```
gslevinkov, line 34: redundant comments, `if` tells us that we check length

gslevinkov (approved):

PR #482, opened 2022-05-16
PWN-3578 - make support of charles and other proxy like apps

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3578

## Description of Work

PWN-3578 - make support of charles and other proxy like apps

### DIFF
--- a/app/src/main/res/xml/network_security_config.xml
+++ b/app/src/main/res/xml/network_security_config.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+    <debug-overrides>
+        <trust-anchors>
+            <certificates src="user" />
+        </trust-anchors>
+    </debug-overrides>
+</network-security-config>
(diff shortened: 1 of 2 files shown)


1sakov (approved):

app/src/main/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+    <debug-overrides>
+        <trust-anchors>
+            <certificates src="user" />
```
gslevinkov, line 5: do you mind adding comment that this certificates are special for Charles?

gslevinkov (approved):

app/src/main/res/xml/network_security_config.xml
kabramovp2p: this is not only for Charles

PR #484, opened 2022-05-16
PWN-3566 - switch to Main tab or back when Send is finished

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3566

## Description of Work

- return to previous screen when send is finished
- or switch to Main tab


app/src/main/java/org/p2p/wallet/send/ui/main/SendFragment.kt
```diff
@@ -124,16 +122,22 @@ class SendFragment :
...
+        childFragmentManager.setFragmentResultListener(KEY_REQUEST_SEND, viewLifecycleOwner) { _, result ->
+            if (result.containsKey(EXTRA_RESULT_KEY_DISMISS)) closeScreenAndReturn()
+        }
+
         presenter.loadInitialData()
         checkClipBoard()
     }
 
+    private fun closeScreenAndReturn() {
+        if (backStackEntryCount() > 1) popBackStack() else replaceFragment(HomeFragment.create())
```
gslevinkov, line 138: is it ok to create HomeFragment here from scratch? maybe i can use bottom tabs to switch to Home??

1sakov (approved):

app/src/main/java/org/p2p/wallet/send/ui/main/SendFragment.kt
1sakov: I think we already have an ability to switch between tabs if you call `popBackStack()`

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/send/ui/main/SendFragment.kt
gslevinkov: popBackStack closes the app itself. I think we are replacing fragments when switching tabs

1sakov: Look into `MainFragment`, there is already implemented switching between tabs when onBackPressed
```
override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        requireActivity().onBackPressedDispatcher.addCallback(this) {
            if (binding.bottomNavigation.selectedItemId != R.id.itemHome) {
                navigate(R.id.itemHome)
                binding.bottomNavigation.menu[0].isChecked = true
            } else {
                requireActivity().finish()
            }
        }
    }
```

gslevinkov: +++

kuznetsov (approved):

davran1001 (approved):

1sakov (approved):

PR #500, opened 2022-05-24
NO-TICKET - Bumped koin version and minor changes

1sakov:
Updated koin version and showed several examples.
Now we can avoid using banch of `get`s in koin module


app/src/main/java/org/p2p/wallet/history/HistoryModule.kt
```diff
@@ -80,12 +81,7 @@ object HistoryModule : InjectionModule {
 
     private fun Module.dataLayer() {
         factory { TransactionDetailsEntityMapper(get()) }
-        single {
-            TransactionDetailsDatabaseRepository(
-                daoDelegate = get(),
-                mapper = get()
-            )
-        } bind TransactionDetailsLocalRepository::class
+        singleOf(::TransactionDetailsDatabaseRepository) bind TransactionDetailsLocalRepository::class
```
gslevinkov, line 84:
```suggestion
        singleOf<TransactionDetailsLocalRepository>(::TransactionDetailsDatabaseRepository)
```

app/src/main/java/org/p2p/wallet/home/HomeModule.kt
```diff
@@ -152,13 +154,7 @@ object HomeModule : InjectionModule {
...
             )
         }
-        factory<TokenListContract.Presenter> {
-            TokenListPresenter(
-                interactor = get(),
-                browseAnalytics = get(),
-                analyticsInteractor = get()
-            )
-        }
+        factoryOf(::TokenListPresenter) bind TokenListContract.Presenter::class
```
gslevinkov, line 157:
```suggestion
        factoryOf<TokenListContract.Presenter>(::TokenListPresenter)
```

app/src/main/java/org/p2p/wallet/home/HomeModule.kt
```diff
@@ -169,15 +165,7 @@ object HomeModule : InjectionModule {
...
-        factory<ReceiveRenBtcContract.Presenter> {
-            ReceiveRenBtcPresenter(
-                interactor = get(),
-                qrCodeInteractor = get(),
-                usernameInteractor = get(),
-                receiveAnalytics = get(),
-                context = get()
-            )
-        }
+        factoryOf(::ReceiveRenBtcPresenter) bind ReceiveRenBtcContract.Presenter::class
```
gslevinkov, line 168:
```suggestion
        factoryOf<ReceiveRenBtcContract.Presenter>(::ReceiveRenBtcPresenter)
```

app/src/main/java/org/p2p/wallet/send/ui/main/SendPresenter.kt
```diff
@@ -699,7 +700,7 @@ class SendPresenter(
         val address = target?.searchAddress?.address
         val isMaxAmount = amount == total
 
-        val isNotZero = !amount.isZero()
+        val isNotZero = amount.isNotZero()
```
gslevinkov, line 703:
```suggestion
        val isAmountNotZero = amount.isNotZero()
```

gslevinkov (approved):

gslevinkov:
Thanks for bumping up the Koin version! Looks great

app/src/main/java/org/p2p/wallet/history/HistoryModule.kt
1sakov: Imho the keyword `bind` looks better than the way you suggest. 
Because it may be confusing which type are you returning
@gslevinkov

kabramovp2p (approved):

davran1001:
ran to read the changes

davran1001 (approved):

app/src/main/java/org/p2p/wallet/history/HistoryModule.kt
kuznetsov: For me both cases are ok but I like Gleb's points. Second way is shorter and looks more idiomatic

kuznetsov (approved):

app/src/main/java/org/p2p/wallet/history/HistoryModule.kt
kabramovp2p: I prefer binds 
but can live with second way :)

davran1001: i prefer <Type> for me it's more clear, cause we already use this arrayListOf<Type> and etc

app/src/main/java/org/p2p/wallet/home/HomeModule.kt
davran1001: here will be the same

app/src/main/java/org/p2p/wallet/history/HistoryModule.kt
1sakov: looks like it's not supported guyz! 
Merging this, and will decide what to do in the future

PR #509, opened 2022-05-31
PWN-3206 - ssl cert added and used in app

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3206

## Description of Work

PWN-3206 - ssl cert added and used in app


app/src/main/java/org/p2p/wallet/infrastructure/network/NetworkModule.kt
```diff
@@ -83,11 +114,26 @@ object NetworkModule : InjectionModule {
...
-    private fun Scope.getClient(tag: String, interceptor: Interceptor? = null): OkHttpClient {
+    private fun Scope.getClient(tag: String, resources: Resources, interceptor: Interceptor? = null): OkHttpClient {
         return OkHttpClient.Builder()
             .readTimeout(DEFAULT_CONNECT_TIMEOUT_SECONDS, TimeUnit.SECONDS)
             .connectTimeout(DEFAULT_READ_TIMEOUT_SECONDS, TimeUnit.SECONDS)
             .apply {
+                val appFeatureFlags: AppFeatureFlags = get()
+                if (appFeatureFlags.isSslPinningEnabled) {
+                    try {
+                        createCertificate(resources.openRawResource(R.raw.cert))?.let { sslContext ->
```
gslevinkov, line 125: I suggest to extract all this pinning logic to a separate entity callled smth like "SslCertificatePinner" and then pass okhttp
it's not Module responsiblity to do such things.

gslevinkov: also, do we need to pin SSL to all the Retrofit instances, or just FeeRelayer?

app/src/main/java/org/p2p/wallet/infrastructure/network/NetworkModule.kt
```diff
@@ -83,11 +114,26 @@ object NetworkModule : InjectionModule {
...
+                            systemDefaultTrustManager()?.let { trustManager ->
+                                Timber.tag(SSL_CERT_TAG).i("SslSocketFactory successfully added with cert")
+                                sslSocketFactory(sslContext.socketFactory, trustManager)
+                            }
+                        }
+                    } catch (e: Exception) {
+                        if (!BuildConfig.DEBUG) {
```
gslevinkov, line 132: Why write this stuff only on prod mode?

app/src/main/java/org/p2p/wallet/infrastructure/network/NetworkModule.kt
```diff
@@ -103,4 +149,43 @@ object NetworkModule : InjectionModule {
             .addNetworkInterceptor(ContentTypeInterceptor())
             .build()
     }
+
+    private fun createCertificate(trustedCertificateIS: InputStream): SSLContext? {
```
gslevinkov, line 153:
```suggestion
    private fun createCertificate(trustedCertificate: InputStream): SSLContext? {
```

you have IS in your type already

app/src/main/java/org/p2p/wallet/infrastructure/network/NetworkModule.kt
```diff
@@ -103,4 +149,43 @@ object NetworkModule : InjectionModule {
...
+            val trustManagers: Array<TrustManager> = trustManagerFactory.trustManagers
+            check(!(trustManagers.size != 1 || trustManagers[0] !is X509TrustManager)) {
+                "Unexpected default trust managers:" + trustManagers.contentToString()
+            }
+            trustManagers[0] as X509TrustManager
+        } catch (e: GeneralSecurityException) {
+            // The system has no TLS. Just give up.
+            throw AssertionError()
+        }
+    }
```
gslevinkov, line 190: you can do an expression body here

app/src/main/java/org/p2p/wallet/infrastructure/network/NetworkModule.kt
```diff
@@ -103,4 +149,43 @@ object NetworkModule : InjectionModule {
...
+            val trustManagerFactory: TrustManagerFactory =
+                TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm())
+            trustManagerFactory.init(null as KeyStore?)
+            val trustManagers: Array<TrustManager> = trustManagerFactory.trustManagers
+            check(!(trustManagers.size != 1 || trustManagers[0] !is X509TrustManager)) {
+                "Unexpected default trust managers:" + trustManagers.contentToString()
+            }
+            trustManagers[0] as X509TrustManager
+        } catch (e: GeneralSecurityException) {
+            // The system has no TLS. Just give up.
```
gslevinkov, line 187: I think we need to log `e` here

app/src/main/java/org/p2p/wallet/infrastructure/network/NetworkModule.kt
```diff
@@ -103,4 +149,43 @@ object NetworkModule : InjectionModule {
...
+        keyStore.load(null, null)
+        keyStore.setCertificateEntry("ca", certificate)
+
+        // creating a TrustManager that trusts the CAs in our KeyStore
+        val trustAlgorithm: String = TrustManagerFactory.getDefaultAlgorithm()
+        val trustManagerFactory: TrustManagerFactory = TrustManagerFactory.getInstance(trustAlgorithm)
+        trustManagerFactory.init(keyStore)
+
+        // creating an SSLSocketFactory that uses our TrustManager
+        val sslContext: SSLContext = SSLContext.getInstance("TLS")
```
gslevinkov, line 171: sonarlint warns (kotlin:S4423):

This rule raises an issue when an insecure TLS protocol version (i.e. a protocol different from "TLSv1.2", "TLSv1.3", "DTLSv1.2", or "DTLSv1.3") is used or allowed.
It is recommended to enforce TLS 1.2 as the minimum protocol version and to disallow older versions like TLS 1.0. Failure to do so could open the door to downgrade attacks: a malicious actor who is able to intercept the connection could modify the requested protocol version and downgrade it to a less secure version.

solution:
```
javax.net.ssl.SSLContext library:
val sc: SSLContext = SSLContext.getInstance("TLSv1.1") // Noncompliant
okhttp library:
val spec: ConnectionSpec = ConnectionSpec.Builder(ConnectionSpec.MODERN_TLS)
                .tlsVersions(TlsVersion.TLS_1_1)  // Noncompliant
                .build()
```

app/src/main/java/org/p2p/wallet/infrastructure/network/NetworkModule.kt
```diff
@@ -83,11 +114,26 @@ object NetworkModule : InjectionModule {
...
+    private fun Scope.getClient(tag: String, resources: Resources, interceptor: Interceptor? = null): OkHttpClient {
         return OkHttpClient.Builder()
             .readTimeout(DEFAULT_CONNECT_TIMEOUT_SECONDS, TimeUnit.SECONDS)
             .connectTimeout(DEFAULT_READ_TIMEOUT_SECONDS, TimeUnit.SECONDS)
             .apply {
+                val appFeatureFlags: AppFeatureFlags = get()
+                if (appFeatureFlags.isSslPinningEnabled) {
+                    try {
+                        createCertificate(resources.openRawResource(R.raw.cert))?.let { sslContext ->
+                            systemDefaultTrustManager()?.let { trustManager ->
```
gslevinkov, line 126: have you tried a newer approach?

https://stackoverflow.com/questions/24006545/how-can-i-pin-a-certificate-with-square-okhttp

I can see your algorithm here is for older (2.x) versions of okhttp

gslevinkov: I can try it, if you want

davran1001 (approved):

app/src/main/java/org/p2p/wallet/infrastructure/network/ssl/CertificateManager.kt
```diff
@@ -0,0 +1,80 @@
...
+import javax.net.ssl.SSLContext
+import javax.net.ssl.TrustManager
+import javax.net.ssl.TrustManagerFactory
+import javax.net.ssl.X509TrustManager
+
+private const val SSL_CERT_TAG = "SSL_CERT"
+
+class CertificateManager(
+    val appFeatureFlags: AppFeatureFlags,
+    val resources: Resources,
```
gslevinkov, line 24:
```suggestion
    private val appFeatureFlags: AppFeatureFlags,
    private val resources: Resources,
```

gslevinkov (approved):

PR #514, opened 2022-06-03
PWN-3611 - Pin code transaction confirmation

kuznetsov:
### Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3611

### Description of Work

- Added `PinView` to `SendConfirmBottomSheet`
- Implemented an ability to confirm transaction by pin code
- User can make unlimited amount of pin code tries

### Screenshots

Screenshot 2022-06-03 at 17 23 49 | Screenshot 2022-06-03 at 17 23 55
:------------:|:-------------:


app/src/main/res/drawable/ic_fingerprint.xml
```diff
@@ -2,10 +2,10 @@
     xmlns:android="http://schemas.android.com/apk/res/android"
     android:width="24dp"
     android:height="24dp"
-    android:tint="@color/backgroundButtonPrimary"
```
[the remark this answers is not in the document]
kuznetsov: Nice point, thanks

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/send/ui/dialogs/SendConfirmBottomSheet.kt
```diff
@@ -82,6 +93,37 @@ class SendConfirmBottomSheet(
...
+    }
+
+    private fun confirm() {
+        onConfirmed.invoke()
+        dismissAllowingStateLoss()
+    }
+
+    private fun checkPinCode(pinCode: String) {
+        CoroutineScope(dispatchers.io).launch {
+            when (authInteractor.signInByPinCode(pinCode)) {
```
davran1001, line 114: why we use interactor inside fragment ?

app/src/main/java/org/p2p/wallet/send/ui/dialogs/SendConfirmBottomSheet.kt
```diff
@@ -82,6 +93,37 @@ class SendConfirmBottomSheet(
...
+        }
+    }
+
+    private fun confirm() {
+        onConfirmed.invoke()
+        dismissAllowingStateLoss()
+    }
+
+    private fun checkPinCode(pinCode: String) {
+        CoroutineScope(dispatchers.io).launch {
```
davran1001, line 113: Redundant use of coroutines, and we should avoid launching not lifecycle aware  coroutines

app/src/main/java/org/p2p/wallet/send/ui/dialogs/SendConfirmBottomSheet.kt
```diff
@@ -82,6 +93,37 @@ class SendConfirmBottomSheet(
...
+            when (authInteractor.signInByPinCode(pinCode)) {
+                SignInResult.Success -> confirm()
+                SignInResult.WrongPin -> withContext(dispatchers.ui) { showWrongPinError() }
+            }
+        }
+    }
+
+    private fun showWrongPinError() {
+        val message = getString(R.string.auth_pin_code_wrong_pin)
+        context?.vibrate(VIBRATE_DURATION)
```
davran1001, line 123: requireContext()

app/src/main/java/org/p2p/wallet/send/ui/dialogs/SendConfirmBottomSheet.kt
```diff
@@ -82,6 +93,37 @@ class SendConfirmBottomSheet(
         }
     }
 
+    private fun initPinView() = with(binding) {
+        pinView.apply {
```
gslevinkov, line 97: too much `this` in scopes, try rewrite using let / also

app/src/main/java/org/p2p/wallet/send/ui/dialogs/SendConfirmBottomSheet.kt
gslevinkov: I think we need to create presenters for BottomSheet now too

app/src/main/java/org/p2p/wallet/utils/BiometricPromptWrapper.kt
```diff
@@ -14,18 +14,17 @@ class BiometricPromptWrapper(
...
 
         override fun onAuthenticationError(errorCode: Int, errString: CharSequence) {
-            onError?.invoke(
-                errString.takeIf {
-                    errorCode != BiometricPrompt.ERROR_USER_CANCELED &&
-                        errorCode != BiometricPrompt.ERROR_NEGATIVE_BUTTON
-                }
-            )
+            when {
+                errorCode == BiometricPrompt.ERROR_NEGATIVE_BUTTON -> usePinCode?.invoke()
```
gslevinkov, line 25: if it's a callback, maybe we should name with with prefix `on`?

gslevinkov (changes requested):

app/src/main/java/org/p2p/wallet/send/ui/dialogs/SendConfirmBottomSheet.kt
kuznetsov: `SendConfirmBottomSheet` is inherited from `NonDraggableBottomSheetDialogFragment`. There are also another bottom sheets inherited from that class and using interactors. I think we can add non draggable logic to `BaseMvpBottomSheet` and get rid of `NonDraggableBottomSheetDialogFragment` without presenters. But it looks like separate refactoring task, so let's do this outside this PR. What do you think, guys?

kuznetsov: https://p2pvalidator.atlassian.net/browse/PWN-3831

davran1001 (approved):

gslevinkov (approved):

PR #516, opened 2022-06-06
PWN-3750 - migrate from resources.getColor -> context.getColor with ResourcesProvider.kt

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3750

## Description of Work

PWN-3750 - migrate from resources.getColor -> context.getColor with ResourcesProvider.kt

### DIFF
--- a/app/src/main/java/org/p2p/wallet/common/ResourcesProvider.kt
+++ b/app/src/main/java/org/p2p/wallet/common/ResourcesProvider.kt
@@ -1,0 +1,17 @@
+package org.p2p.wallet.common
+
+import android.content.Context
+import android.content.res.Resources
+import androidx.annotation.ColorRes
+import androidx.annotation.StringRes
+
+class ResourcesProvider(private val context: Context) {
+
+    val resources: Resources = context.resources
+
+    fun getColor(@ColorRes color: Int) = context.getColor(color)
+
+    fun getString(@StringRes stringRes: Int) = context.getString(stringRes)
+
+    fun getString(@StringRes stringRes: Int, vararg formatArgs: Any) = context.getString(stringRes, *formatArgs)
+}
--- a/app/src/main/java/org/p2p/wallet/common/ResourcesProvider.kt
+++ b/app/src/main/java/org/p2p/wallet/common/ResourcesProvider.kt
@@ -13,5 +13,9 @@ class ResourcesProvider(private val context: Context) {
 
     fun getString(@StringRes stringRes: Int) = context.getString(stringRes)
 
-    fun getString(@StringRes stringRes: Int, vararg formatArgs: Any) = context.getString(stringRes, *formatArgs)
+    fun getString(@StringRes stringRes: Int, vararg formatArgs: Any) = if (formatArgs.isNotEmpty()) {
+        context.getString(stringRes, *formatArgs)
+    } else {
+        getString(stringRes)
+    }
 }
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/p2p/wallet/common/mvp/BaseFragment.kt
```diff
@@ -77,7 +77,7 @@ abstract class BaseFragment(@LayoutRes layoutRes: Int) : Fragment(layoutRes), Ba
         val window = requireActivity().window ?: return
         with(window) {
             addFlags(WindowManager.LayoutParams.FLAG_DRAWS_SYSTEM_BAR_BACKGROUNDS)
-            statusBarColor = resources.getColor(colorResId, requireActivity().theme)
+            statusBarColor = requireContext().getColor(colorResId)
```
kuznetsov, line 80: Why didn't you inject `ResourcesProvider` and use here?

kuznetsov (changes requested):

app/src/main/java/org/p2p/wallet/common/mvp/BaseFragment.kt
kabramovp2p: i thought use it only in presenters or interractors

but on fragment side use context itself

kuznetsov: Got it)

kuznetsov (approved):

app/src/main/java/org/p2p/wallet/common/mvp/BaseFragment.kt
davran1001: It's bad idea to have 2 ways to get color, if you create some provider class, so we need to use it everywhere

app/src/main/java/org/p2p/wallet/common/ResourcesProvider.kt
```diff
@@ -0,0 +1,17 @@
+package org.p2p.wallet.common
+
+import android.content.Context
+import android.content.res.Resources
+import androidx.annotation.ColorRes
+import androidx.annotation.StringRes
+
+class ResourcesProvider(private val context: Context) {
```
davran1001, line 8: I think this class is redundant, cause we can have just extensions, i guess we already them. So which problems resolve this class ?

davran1001:
confused way to provide resources )), what do you think, will not it be overwhelmed to replace it everywhere

app/src/main/java/org/p2p/wallet/common/ResourcesProvider.kt
gslevinkov: mocks / stubs in unit tests?

app/src/main/java/org/p2p/wallet/common/ResourcesProvider.kt
```diff
@@ -0,0 +1,17 @@
...
+
+    val resources: Resources = context.resources
+
+    fun getColor(@ColorRes color: Int) = context.getColor(color)
+
+    fun getString(@StringRes stringRes: Int) = context.getString(stringRes)
+
+    fun getString(@StringRes stringRes: Int, vararg formatArgs: Any) = context.getString(stringRes, *formatArgs)
```
gslevinkov, line 16: I can suggest to check formatArgs for emptiness
```
if (formatArgs.isNotEmpty()) {
        context.getString(stringId, *formatArgs)
    } else {
        context.getString(stringId)
    }
```

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/common/mvp/BaseFragment.kt
kabramovp2p: ok will apply to baseFragment also

app/src/main/java/org/p2p/wallet/common/ResourcesProvider.kt
1sakov: In some cases we need resources in presenters, but our extensions are expected to be used for Views. 
Thus, there is no other way that including context/resources/resources provider in presenter. 

The only thing I was confused is injecting the theme because getting colors via resources requires Theme right now. 
We decided to find some way, where we don't think about theme injection.

PR #518, opened 2022-06-06
PWN-3724 - Fix SOL price on Derivable Accounts

kuznetsov:
### Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3724

### Description of Work

Since token rates are requested on `HomeFragment` we don't have it on `DerivableAccountsFragment`. I requested `SOL` price only on `DerivableAccountsFragment` to calculate total USD balances.

### DIFF
--- a/app/src/main/java/org/p2p/wallet/restore/interactor/SecretKeyInteractor.kt
+++ b/app/src/main/java/org/p2p/wallet/restore/interactor/SecretKeyInteractor.kt
@@ -15,12 +15,13 @@ import org.p2p.wallet.restore.model.DerivableAccount
 import org.p2p.wallet.restore.model.SecretKey
 import org.p2p.wallet.restore.model.SeedPhraseResult
 import org.p2p.wallet.rpc.repository.balance.RpcBalanceRepository
-import org.p2p.wallet.user.repository.UserLocalRepository
+import org.p2p.wallet.user.repository.prices.TokenPricesRemoteRepository
+import org.p2p.wallet.user.repository.prices.TokenSymbol
 import org.p2p.wallet.utils.Constants.SOL_SYMBOL
+import org.p2p.wallet.utils.Constants.USD_READABLE_SYMBOL
 import org.p2p.wallet.utils.fromLamports
 import org.p2p.wallet.utils.mnemoticgenerator.English
 import org.p2p.wallet.utils.scaleLong
-import java.math.BigDecimal
 import java.math.BigInteger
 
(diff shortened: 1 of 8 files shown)


kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/user/repository/prices/impl/TokenPricesCoinGeckoRepository.kt
```diff
@@ -36,6 +36,13 @@ class TokenPricesCoinGeckoRepository(
         loadPrices(tokenSymbols, targetCurrency)
     }
 
+    override suspend fun getTokenPriceBySymbol(
+        tokenSymbol: TokenSymbol,
+        targetCurrency: String
+    ): TokenPrice {
+        return loadPrices(listOf(tokenSymbol), targetCurrency).first()
```
1sakov, line 43: Are we able to load a price for a single token?

kuznetsov: Actually there are methods for single prices, I just used the implemented ones

kuznetsov: Actually, you're right. If we're sending request for one token price, why not requesting all and cache them before home

kuznetsov: But if we want to load and cache all user tokens prices on DerivableAccount we should make 2 bigger requests and make caching job, screen could be overloaded

app/src/main/java/org/p2p/wallet/restore/interactor/SecretKeyInteractor.kt
```diff
@@ -56,13 +60,18 @@ class SecretKeyInteractor(
             return@withContext result
         }
 
-    private fun mapDerivableAccounts(
+    private suspend fun mapDerivableAccounts(
         accounts: List<Account>,
         balances: List<Pair<String, BigInteger>>,
         path: DerivationPath
     ) = accounts.mapNotNull { account ->
```
gslevinkov, line 67:
```suggestion
    ) : List< DerivableAccount> = accounts.mapNotNull { account ->
```

gslevinkov (approved):

PR #534, opened 2022-06-14
PWN-3775 - Fix restore key button state

davran1001:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3775

### DIFF
--- a/app/src/main/java/org/p2p/wallet/restore/ui/keys/SecretKeyFragment.kt
+++ b/app/src/main/java/org/p2p/wallet/restore/ui/keys/SecretKeyFragment.kt
@@ -41,6 +41,9 @@ import org.p2p.wallet.utils.viewbinding.viewBinding
 import timber.log.Timber
 import java.io.File
 
+private const val SEED_PHRASE_SIZE_SHORT = 12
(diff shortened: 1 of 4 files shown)


gslevinkov (approved):

app/src/main/java/org/p2p/wallet/restore/ui/keys/SecretKeyFragment.kt
```diff
@@ -86,8 +86,6 @@ class SecretKeyFragment :
             termsAndConditionsTextView.text = buildTermsAndPrivacyText()
             termsAndConditionsTextView.movementMethod = LinkMovementMethod.getInstance()
         }
-
-        setButtonEnabled(phraseAdapter.itemCount != 0)
```
1sakov, line 90: We need to enable the button if we navigated back to this screen. 
I think we shouldn't delete it but add check for `phraseAdapter.itemCount == 12 or 24`

1sakov (changes requested):

kabramovp2p (approved):

kuznetsov (approved):

1sakov (approved):

PR #537, opened 2022-06-15
PWN-3100 - Started working on unit tests for SendPresenter

1sakov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3100

## Description of Work

Some of the classes required changes. 
For example, dispatchers. For testing functions with coroutines we need to replace the real dispatcher to the TestDispatcher, that's why, we need to inject `CoroutineDispatchers` all the time. 

FvSax


app/src/main/java/org/p2p/wallet/send/ui/main/SendPresenter.kt
```diff
@@ -102,9 +102,10 @@ class SendPresenter(
     }
 
     override fun loadInitialData() {
-        launch {
+        launch(dispatchers.ui) {
```
1sakov, line 105: The only way, I've found to replace the real dispatcher to TestDispatcher

davran1001 (approved):

PR #539, opened 2022-06-16
PWN-3595 - save images with timestamp to make not override latest images of QRCodes

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3595

## Description of Work

Save images with timestamp to make not override latest images of QRCodes

### DIFF
--- a/app/src/main/java/org/p2p/wallet/auth/ui/security/SecurityKeyPresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/auth/ui/security/SecurityKeyPresenter.kt
@@ -9,7 +9,7 @@ import org.p2p.wallet.common.mvp.BasePresenter
 import org.p2p.wallet.restore.interactor.SecretKeyInteractor
 import kotlin.properties.Delegates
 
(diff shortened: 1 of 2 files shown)


davran1001 (approved):

app/src/main/java/org/p2p/wallet/auth/ui/security/SecurityKeyPresenter.kt
```diff
@@ -9,6 +9,8 @@ import org.p2p.wallet.common.mvp.BasePresenter
 import org.p2p.wallet.restore.interactor.SecretKeyInteractor
 import kotlin.properties.Delegates
 
+private const val SECURE_KEY_NAME = "SecureKey"
```
gslevinkov, line 12:
```suggestion
private const val SECURITY_KEY_IMAGE_NAME = "SecureKey"
```

gslevinkov (approved):

kuznetsov (approved):

PR #544, opened 2022-06-17
PWN-3681 - UiKit Bottom Navigation View

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3681

## Description of Work

- new, ready for use, bottom navigation view

## Screenshots (Optional)
image


settings.gradle
```diff
@@ -1,4 +1,5 @@
 include ':solanaj'
 include ':app'
 include ':ui-kit'
-rootProject.name = "P2P Wallet"
\ No newline at end of file
+rootProject.name = "P2P Wallet"
+include ':common'
```
gslevinkov, line 5:
```suggestion
include ':common'
rootProject.name = "P2P Wallet"
```

ui-kit/src/main/res/values/typography.xml
```diff
@@ -1,120 +1,118 @@
...
-    <style name="TextAppearance.Regular.LargeTitle" parent="TextAppearance.Regular">
+    <style name="UiKit.TextAppearance.Regular.LargeTitle" parent="UitKit_TextAppearance.Regular">
         <item name="android:textSize">34sp</item>
         <item name="android:letterSpacing">0.01</item>
     </style>
     <!--endregion-->
 
     <!--region TextAppearance.SemiBold-->
-    <style name="TextAppearance.SemiBold">
+    <style name="UitKit_TextAppearance.SemiBold">
```
kabramovp2p, line 65: why UitKit_TextAppearance and then UiKit.TextAppearance ?

davran1001 (approved):

ui-kit/src/main/res/values/typography.xml
gslevinkov: you mean why UitKit_TextAppearance and not UiKit.TextAppearance?

1sakov:
no way! 
Why did you create a common module?

gslevinkov:
> no way! Why did you create a common module?

we need a place where we can put extensions to use both in :ui-kit and :app module without needing to duplicate code

1sakov:
> > no way! Why did you create a common module?
> 
> we need a place where we can put extensions to use both in :ui-kit and :app module without needing to duplicate code

why don't we put these extensions in ui-kit? 
We are using all ui stuff from this module anyway

gslevinkov:
> > > no way! Why did you create a common module?
> > 
> > 
> > we need a place where we can put extensions to use both in :ui-kit and :app module without needing to duplicate code
> 
> why don't we put these extensions in ui-kit? We are using all ui stuff from this module anyway.
> 
> Splitting for common module is a big process. We need to discuss first, what should we move to common module, how to call it, what rules should we follow if we start with common modules

sounds reasonable enough, i'll do a revert

ui-kit/src/main/res/values/typography.xml
kabramovp2p: yes! =)

kabramovp2p (approved):

ui-kit/src/main/res/values/typography.xml
gslevinkov: check again please, is it ok for you?

PR #551, opened 2022-06-22
PWN-3806 - Add feature toggles in app

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3806

## Description of Work

- add feature toggle entity with subclasses that declare concrete types
- edit feature toggles settings screen
- add RemoteConfig fetch from Firebase

## Screenshots (Optional)
Screen Shot 2022-06-22 at 15 24 54


app/src/main/java/org/p2p/wallet/debug/feature_toggles/FeatureTogglesAdapter.kt
```diff
@@ -0,0 +1,102 @@
...
+            return oltItem.toggleName == newItem.toggleName && oltItem.toggleValue == newItem.toggleValue
+        }
+
+        override fun areContentsTheSame(oldItemPosition: Int, newItemPosition: Int): Boolean {
+            return oldList[oldItemPosition] == newList[newItemPosition]
+        }
+    }
+
+    // not so much data for using DiffUtil
+    @SuppressLint("NotifyDataSetChanged")
```
kabramovp2p, line 43: do we need this supress?

app/src/main/java/org/p2p/wallet/debug/settings/DebugSettingsFragment.kt
```diff
@@ -7,7 +7,7 @@ import org.koin.android.ext.android.inject
 import org.p2p.wallet.R
 import org.p2p.wallet.common.mvp.BaseMvpFragment
 import org.p2p.wallet.databinding.FragmentDebugSettingsBinding
-import org.p2p.wallet.debug.featuretoggles.FeatureTogglesFragment
```
kabramovp2p, line 10: do not like _ in packages

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/debug/settings/DebugSettingsFragment.kt
gslevinkov: hm, i thought we have set naming in our code style, but i was wrong.
should we use Kotlin `camelCase.packageName` style then?

kabramovp2p: @1sakov ^^ ?
i prefer camelCase if this won't harm our CI for example

1sakov: I'd leave as it is. 
We have several similar folders already. like pushnotifications

So, let's leave as `featuretoggles`

1sakov (changes requested):
Added comments. 
In general LGTM, it's extendable when new feature is added and I like it. 

But I'm confused what class should I use when I want to hide it behind feature toggle? 
There are different classes can be injected and imho this is bad

app/src/main/java/org/p2p/wallet/common/feature_toggles/FeatureToggle.kt
```diff
@@ -0,0 +1,49 @@
...
+abstract class FloatFeatureToggle(
+    private val valuesProvider: RemoteConfigValuesSource
+) : FeatureToggle<Float>() {
+    override val value: Float get() = valuesProvider.getFloat(toggleKey) ?: defaultValue
+}
+
+/**
+ * Use subclasses to override how should be JSON parsed
+ */
+abstract class JsonFeatureToggle<GSON_CLASS>(
```
davran1001, line 47: may name it's as GsonInstance or GsonClass

davran1001 (approved):

app/src/main/java/org/p2p/wallet/debug/settings/DebugSettingsFragment.kt
gslevinkov: done

app/src/main/java/org/p2p/wallet/debug/feature_toggles/FeatureTogglesAdapter.kt
gslevinkov: removed

app/src/main/java/org/p2p/wallet/debug/feature_toggles/FeatureTogglesAdapter.kt
```diff
@@ -0,0 +1,102 @@
...
+
+        lateinit var currentRow: FeatureToggleRow
+
+        private val toggleValueInputWatcher = object : TextWatcher {
+            override fun onTextChanged(text: CharSequence?, start: Int, before: Int, count: Int) {
+                text?.let { onToggleChanged.invoke(currentRow, it.toString()) }
+            }
+
+            override fun afterTextChanged(s: Editable?) = NoOp
+            override fun beforeTextChanged(s: CharSequence?, start: Int, count: Int, after: Int) = NoOp
```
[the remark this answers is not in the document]
gslevinkov: Let's discuss this on the sync

1sakov (approved):

gslevinkov:
check hook

kuznetsov (approved):

PR #558, opened 2022-06-27
PWN-3974 - fix token History by reloading if no transactions

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3974

## Description of Work
Fix token History by reloading if no transactions

### DIFF
--- a/app/src/main/java/org/p2p/wallet/history/ui/token/TokenHistoryPresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/history/ui/token/TokenHistoryPresenter.kt
@@ -94,10 +94,7 @@ class TokenHistoryPresenter(
             if (isRefresh) {
                 transactions.clear()
             }
-            val fetchedItems = historyInteractor.loadTransactions(
-                token.publicKey,
-                isRefresh = isRefresh || transactions.isEmpty()
-            )
+            val fetchedItems = historyInteractor.loadTransactions(token.publicKey)
             transactions.addAll(fetchedItems)
             view?.showHistory(transactions)
             view?.showPagingState(PagingState.Idle)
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/p2p/wallet/history/ui/token/TokenHistoryPresenter.kt
```diff
@@ -94,7 +94,10 @@ class TokenHistoryPresenter(
             if (isRefresh) {
                 transactions.clear()
             }
-            val fetchedItems = historyInteractor.loadTransactions(token.publicKey, isRefresh)
+            val fetchedItems = historyInteractor.loadTransactions(
+                token.publicKey,
+                isRefresh = isRefresh || transactions.isEmpty()
```
1sakov, line 99: I am not sure if it's a best solution. 
cannot get how this field effects the empty state

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/history/ui/token/TokenHistoryPresenter.kt
kabramovp2p: yes i agree
i'm consulting with @davran1001 
but it looks like issue in caches of states
and reloading fixes it

i will spend more time to check the root cause of this issue

1sakov: then it's wip?

kabramovp2p: fixed @1sakov

davran1001 (approved):

1sakov (approved):

PR #572, opened 2022-07-02
PWN-3891 - UI Kit Cells Atoms

kuznetsov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/pwn-3891

## Description of Work

Implemented following cells atoms:

- `UiKitBadgeView`
- `UiKitCounterView`
- `UiKitLeftAmountView`
- `UiKitRightAmountView`

## Screenshots

Screenshot 2022-07-02 at 15 30 20


kuznetsov:
Old way of `ContextCompat` methods were used. Should be replaced with utils

kabramovp2p (approved):
looks good to me

I could write some small changes but it is unnecessary... 👍🏻

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitBadgeView.kt
```diff
@@ -0,0 +1,92 @@
...
+            OVAL,
+            RECTANGLE
+        }
+    }
+
+    init {
+        context.obtainStyledAttributes(attrs, R.styleable.UiKitBadgeView).use { typedArray ->
+            val defaultTextColor = ContextCompat.getColor(context, R.color.mountain)
+            val hasStroke = typedArray.getBoolean(R.styleable.UiKitBadgeView_hasStroke, false)
+            val shape = when (typedArray.getInt(R.styleable.UiKitBadgeView_shape, 0)) {
```
gslevinkov, line 35: if

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitBadgeView.kt
```diff
@@ -0,0 +1,92 @@
...
+            VERTICAL_PADDING_DP.toPx()
+        )
+    }
+
+    private fun setBackground(shape: Shape, hasStroke: Boolean) {
+        val ovalCornerRadius = resources.getDimension(R.dimen.ui_kit_badge_oval_corner_radius)
+        val rectangleCornerRadius = resources.getDimension(R.dimen.ui_kit_badge_rectangle_corner_radius)
+        val cornerRadiusDp = if (shape == Shape.OVAL) ovalCornerRadius else rectangleCornerRadius
+        val strokeBackgroundId = if (shape == Shape.OVAL) {
+            R.drawable.backgroun_badge_stroke_oval
```
gslevinkov, line 59: backgroun_badge_stroke_oval -> background_....

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitCounterView.kt
```diff
@@ -0,0 +1,80 @@
...
+            HORIZONTAL_PADDING_DP.toPx(),
+            VERTICAL_PADDING_DP.toPx(),
+            HORIZONTAL_PADDING_DP.toPx(),
+            VERTICAL_PADDING_DP.toPx()
+        )
+    }
+
+    override fun setText(text: CharSequence?, type: BufferType?) {
+        when {
+            text.isNullOrBlank() -> count = 0
```
gslevinkov, line 54: add or remove brackets to all

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitCounterView.kt
```diff
@@ -0,0 +1,80 @@
...
+            !text.all { it.isDigit() || it == PLUS_SIGN } -> {
+                throw IllegalArgumentException("Text should be digits and '+' only")
+            }
+            else -> super.setText(text, type)
+        }
+    }
+
+    fun inc() {
```
gslevinkov, line 62: i don't think that you can increment the view itself. how bout `incCounter`?

gslevinkov (approved):

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitBadgeView.kt
```diff
@@ -0,0 +1,91 @@
...
+    enum class Shape {
+        OVAL,
+        RECTANGLE
+    }
+
+    init {
+        context.obtainStyledAttributes(attrs, R.styleable.UiKitBadgeView).use { typedArray ->
+            val defaultTextColor = ContextCompat.getColor(context, R.color.mountain)
+            val hasStroke = typedArray.getBoolean(R.styleable.UiKitBadgeView_hasStroke, false)
+            val shape = if (typedArray.getInt(R.styleable.UiKitBadgeView_shape, 0) == 0) {
```
1sakov, line 35: Why you didn't make it as text? 
It would be easy to parsing, for example:
`Shape.valueOf(shape)` and you won't need `if` after it

Moreover, 1 or 0 is confusing, the developer doesn't no if 0 is `Oval` and 1 is `Rectangle`

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitBadgeView.kt
```diff
@@ -0,0 +1,91 @@
...
+            VERTICAL_PADDING_DP.toPx()
+        )
+    }
+
+    private fun setBackground(shape: Shape, hasStroke: Boolean) {
+        val ovalCornerRadius = resources.getDimension(R.dimen.ui_kit_badge_oval_corner_radius)
+        val rectangleCornerRadius = resources.getDimension(R.dimen.ui_kit_badge_rectangle_corner_radius)
+        val cornerRadiusDp = if (shape == Shape.OVAL) ovalCornerRadius else rectangleCornerRadius
+        val strokeBackgroundId = if (shape == Shape.OVAL) {
+            R.drawable.background_badge_stroke_oval
```
1sakov, line 60: Maybe we need to move this variables to enum class? 
You'll be able to get it easily by `shape.backgroundRes` for example

You can make the same to radius as well

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitBadgeView.kt
```diff
@@ -0,0 +1,91 @@
...
+    private fun getBackgroundColor(): Int {
+        return (background as? ColorDrawable)?.color ?: ContextCompat.getColor(context, R.color.lime)
+    }
+
+    private fun createFilledBackground(
+        backgroundColor: Int,
+        cornerRadiusDp: Float,
+    ): LayerDrawable {
+        val cornerRadiusPx = cornerRadiusDp.toPx()
+        val corners = FloatArray(RADIUS_LIST_SIZE) { cornerRadiusPx }
```
1sakov, line 79: Why `RADIUS_LIST_SIZE` is 8 not 4?

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitBadgeView.kt
```diff
@@ -0,0 +1,91 @@
...
+        val filledBackground = createFilledBackground(backgroundColor, cornerRadiusDp)
+        background = if (hasStroke) strokeBackground else filledBackground
+    }
+
+    private fun getBackgroundColor(): Int {
+        return (background as? ColorDrawable)?.color ?: ContextCompat.getColor(context, R.color.lime)
+    }
+
+    private fun createFilledBackground(
+        backgroundColor: Int,
```
1sakov, line 75: Please use annotation for all types. `@ColorInt`, `@DimenRes` etc.

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitBadgeView.kt
```diff
@@ -0,0 +1,91 @@
...
+) : AppCompatTextView(context, attrs, defStyleAttr) {
+
+    enum class Shape {
+        OVAL,
+        RECTANGLE
+    }
+
+    init {
+        context.obtainStyledAttributes(attrs, R.styleable.UiKitBadgeView).use { typedArray ->
+            val defaultTextColor = ContextCompat.getColor(context, R.color.mountain)
```
1sakov, line 33: Please use native method `context.getColor`
It uses compat for lower OS versions and standard get color for newer versions automatically

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitCounterView.kt
```diff
@@ -0,0 +1,80 @@
...
+            VERTICAL_PADDING_DP.toPx()
+        )
+    }
+
+    override fun setText(text: CharSequence?, type: BufferType?) {
+        val errorMessage = "Text should be digits and '+' only"
+
+        when {
+            text.isNullOrBlank() -> count = 0
+            !text.all { it.isDigit() || it == PLUS_SIGN } -> throw IllegalArgumentException(errorMessage)
```
1sakov, line 57: We shouldn't crash our app. 
Imho we just should show empty string if there are letters only. 

And maybe it'd be easier to read if we filter it this way: 
`text.none { it.isDigit() || it == PLUS_SIGN }`

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitCounterView.kt
```diff
@@ -0,0 +1,80 @@
...
+private const val HORIZONTAL_PADDING_DP = 4
+private const val PLUS_SIGN = '+'
+
+class UiKitCounterView @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : AppCompatTextView(context, attrs, defStyleAttr) {
+
+    var count = 0
```
1sakov, line 23: I don't really like this patter here. 
I think the components should have only functions.

`setCount`, `updateCount`

The same should be in all other views

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitCounterView.kt
```diff
@@ -0,0 +1,80 @@
...
+        set(value) {
+            text = validateCount(value)
+            field = value
+        }
+
+    var isRead: Boolean = false
+        set(value) {
+            setBackground(value)
```
1sakov, line 31: What is `isRead`? 
And why we don't set the color just from xml params?

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitLeftAmountView.kt
```diff
@@ -0,0 +1,61 @@
...
+import org.p2p.uikit.utils.withTextOrGone
+
+class UiKitLeftAmountView @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : ConstraintLayout(context, attrs, defStyleAttr) {
+
+    @DrawableRes
+    var icon: Int? = null
```
1sakov, line 21: same here, please make it functions and give appropriate names

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitLeftAmountView.kt
```diff
@@ -0,0 +1,61 @@
...
+            field = value
+        }
+
+    var subtitle1: String? = null
+        set(value) {
+            binding.subtitle1TextView.withTextOrGone(value)
+            field = value
+        }
+
+    var subtitle2: String? = null
```
1sakov, line 39: Maybe we could avoid using numbers in ids?

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitLeftAmountView.kt
```diff
@@ -0,0 +1,61 @@
...
+            binding.subtitle2TextView.withTextOrGone(value)
+            field = value
+        }
+
+    private val binding = inflateViewBinding<WidgetLeftAmountViewBinding>()
+
+    init {
+        context.obtainStyledAttributes(attrs, R.styleable.UiKitLeftAmountView).use { typedArray ->
+            val iconRes = typedArray.getResourceId(R.styleable.UiKitLeftAmountView_icon, -1)
+            val defaultSubtitleColor = ContextCompat.getColor(context, R.color.text_mountain)
```
1sakov, line 50: Use `context.getColor`

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitLeftAmountView.kt
```diff
@@ -0,0 +1,61 @@
...
+import androidx.annotation.DrawableRes
+import androidx.constraintlayout.widget.ConstraintLayout
+import androidx.core.content.ContextCompat
+import org.p2p.uikit.R
+import org.p2p.uikit.databinding.WidgetLeftAmountViewBinding
+import org.p2p.uikit.utils.inflateViewBinding
+import org.p2p.uikit.utils.withImageOrGone
+import org.p2p.uikit.utils.withTextOrGone
+
+class UiKitLeftAmountView @JvmOverloads constructor(
```
1sakov, line 14: Let's split it of `UiKitStartAmountView` and `UiKitEndAmountView`

The current name is confusing a little bit, because the developer can assume that it's about some remaining value, not about the position

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitLeftAmountView.kt
```diff
@@ -0,0 +1,61 @@
...
+        set(value) {
+            binding.subtitle2TextView.withTextOrGone(value)
+            field = value
+        }
+
+    private val binding = inflateViewBinding<WidgetLeftAmountViewBinding>()
+
+    init {
+        context.obtainStyledAttributes(attrs, R.styleable.UiKitLeftAmountView).use { typedArray ->
+            val iconRes = typedArray.getResourceId(R.styleable.UiKitLeftAmountView_icon, -1)
```
1sakov, line 49: this should work, you'll not need to add an additional `if` condition later
```suggestion
            icon = typedArray.getResourceId(R.styleable.UiKitLeftAmountView_icon, 0)
```

ui-kit/src/main/res/drawable/background_counter_new.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle">
+
+    <solid android:color="@color/night" />
+    <corners android:radius="10dp" />
```
1sakov, line 7: Please, check one more time, we shouldn't have such radiuses. 
it should be 8 or 12

ui-kit/src/main/res/drawable/ic_info_outline.xml
```diff
@@ -0,0 +1,11 @@
+<vector
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+    <path
+        android:pathData="M12,2C6.48,2 2,6.48 2,12C2,17.52 6.48,22 12,22C17.52,22 22,17.52 22,12C22,6.48 17.52,2 12,2ZM11,7V9H13V7H11ZM13,16C13,16.55 12.55,17 12,17C11.45,17 11,16.55 11,16V12C11,11.45 11.45,11 12,11C12.55,11 13,11.45 13,12V16ZM4,12C4,16.41 7.59,20 12,20C16.41,20 20,16.41 20,12C20,7.59 16.41,4 12,4C7.59,4 4,7.59 4,12Z"
+        android:fillColor="#6F7D8D"
```
1sakov, line 9: use color from theme

ui-kit/src/main/res/layout/widget_left_amount_view.xml
```diff
@@ -0,0 +1,53 @@
...
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    tools:parentTag="androidx.constraintlayout.widget.ConstraintLayout">
+
+    <ImageView
+        android:id="@+id/imageView"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:paddingEnd="12dp"
```
1sakov, line 14: it's better to use margin instead of padding. 
Padding requires RTL balance

1sakov (changes requested):

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitLeftAmountView.kt
1sakov: But in general, these two views are absolutely different and perhaps we shouldn't make their names similar dividing to start/end or left/right. 

These two components have different text views and images so let's think about more informative names

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitBadgeView.kt
kuznetsov: This attribute has limited values so for easy using it in xml it should be enum. But making custom `valueOf()` function is a nice point

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitBadgeView.kt
kuznetsov: Then it's better to call `RADIUSES_LIST_SIZE`

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitBadgeView.kt
kuznetsov: You're right, `context.getColor()` is preferred in our case. But you're a bit wrong about method, cause `context.getColor()` doesn't make any compatibility checks and `ContextCompat.getColor()` does. But it useless for us, cause it uses native method for 23+ SDK versions and we have 26 as minimum anyway

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitLeftAmountView.kt
kuznetsov: Yes, it's way better! I was confused about naming this guys

ui-kit/src/main/res/layout/widget_left_amount_view.xml
kuznetsov: In that case margin is ignored by ConstraintLayout for some reason

ui-kit/src/main/res/drawable/background_counter_new.xml
kuznetsov: ✔

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitCounterView.kt
kuznetsov: Agree about crashing the app. But this view could have only digits and `+`, no empty string by definition, let's keep `0` as default value.
`text.none { it.isDigit() || it == PLUS_SIGN }` – this is different predicate

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitCounterView.kt
kuznetsov: In Kotlin style we usually call `textView.text =` or `view.graviti =` , `view.isVisible = `, etc. Property style is pretty common

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitCounterView.kt
kuznetsov: I wanted to encapsulate this two colours in view, because counter can be only one of them. But I'm ok with replacing it with usual background if you want so

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitLeftAmountView.kt
kuznetsov: Let's name it `subtitle` and `subSubtitle`. I can't come up with better naming 🤷‍♂️

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitCounterView.kt
1sakov: Okay with colors. 
What is `isRead`? 
What does it mean?

1sakov (approved):

ui-kit/src/main/kotlin/org/p2p/uikit/atoms/UiKitCounterView.kt
gslevinkov: maybe `isBadgeVisible`?

PR #577, opened 2022-07-05
PWN-4069 - Push notification service url

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-4069

## Description of Work

Screen for push notification service url change added to debug settings


app/src/main/res/values/environment_config.xml
```diff
@@ -25,5 +25,6 @@
 
     <string name="orca_api_base_url">https://api.orca.so/</string>
 
-    <string name="notification_service_url">https://push-service.wallet.p2p.org/</string>
+    <string name="notificationServiceBaseUrl">https://push-service.wallet.p2p.org/</string>
+    <string name="notificationServiceTestBaseUrl">http://35.234.120.240:9090/</string>
```
1sakov, line 29: is it working with http? 
Are you sure we don't need to add network config?

kabramovp2p: network config ?

kabramovp2p: it should works as far as QA checked 

also it was done same way lire feerelayer url changes 

but we can do it in some other way 

just let me know

1sakov: yes, if I remember correctly we should have issues with `http` requests, not `https`. 

For such cases, we had to use `useClearTextTraffic` or smth like that, or add network_config.xml and include it inside the manifest

kabramovp2p: hmmm... ok I will add this address to config but it is strange that it was ok on testing :(

kabramovp2p: @1sakov  done

1sakov (approved):

davran1001 (approved):

PR #604, opened 2022-07-20
PWN-4052 - Add UI KIT module

davran1001:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-4052

Add submodule of UI KIT module
Screen Shot 2022-07-20 at 19 13 10
Screen Shot 2022-07-20 at 19 13 22
Screen Shot 2022-07-20 at 19 13 43


app/build.gradle
```diff
@@ -201,6 +207,7 @@ android {
 dependencies {
     implementation project(':solanaj')
     implementation project(':ui-kit')
+    implementation project(':ui-kit-sample')
```
kabramovp2p, line 210: it seems to me that app have all dependencies
can you check size of app with this impl and without ?

i do not like it if we will have more then we need in our app like size (classes / resources/ etc)

app/src/main/AndroidManifest.xml
```diff
@@ -50,6 +51,21 @@
 
         </activity>
 
+        <activity
```
kabramovp2p, line 54: yes and same here
if we will install this app there will be 2 icons for run app and sample
as far as i see

app/build.gradle
davran1001: 33 MB vs 32 MB

gslevinkov:
fix the CircleCI please

settings.gradle
```diff
@@ -1,4 +1,19 @@
 include ':solanaj'
 include ':app'
-include ':ui-kit'
-rootProject.name = "P2P Wallet"
\ No newline at end of file
+rootProject.name = "P2P Wallet"
+addSubfolderProjects("p2p-wallet-ui-app/ui-kit")
```
kuznetsov, line 4: I don't think the `p2p-wallet-ui-app` is a good name for that repo. It is actually ui kit repository with ui kit library itself with the sample

kuznetsov (changes requested):

settings.gradle
kuznetsov: May be the `android-ui-kit`

PR #609, opened 2022-07-22
PWN-4263 - start and continue screens added

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-4263

## Description of Work

Start and continue screens added

## Screenshots (Optional)

Screenshot_20220722_150854
Screenshot_20220722_150953


app/src/main/java/org/p2p/wallet/auth/common/WalletWeb3AuthManager.kt
```diff
@@ -93,15 +94,28 @@ class WalletWeb3AuthManager(
...
+        }
+    }
+
+    private fun parseDeviceShare(deviceShare: String): DeviceShareKey? {
+        return gson.fromJson(deviceShare, DeviceShareKey::class.java)
     }
 
-    private fun hasDeviceShare(): Boolean = with(sharedPreferences) {
-        contains("${KEY_DEVICE_SHARE}_$lastUserId")
+    fun hasDeviceShare(): Boolean = with(sharedPreferences) {
```
gslevinkov, line 114: why `with`?

app/src/main/java/org/p2p/wallet/auth/common/WalletWeb3AuthManager.kt
```diff
@@ -93,15 +94,28 @@ class WalletWeb3AuthManager(
...
+        contains(KEY_DEVICE_SHARE)
     }
 
-    private fun getDeviceShare(): String = secureStorage.getString("${KEY_DEVICE_SHARE}_$lastUserId").orEmpty()
+    private fun getDeviceShare(): String = secureStorage.getString(KEY_DEVICE_SHARE).orEmpty()
```
gslevinkov, line 118: are you sure we want to return empty string in case of absent device share?
I think it's okay to return `null` if no device share is found

app/src/main/java/org/p2p/wallet/auth/model/DeviceShareKey.kt
```diff
@@ -6,6 +6,7 @@ data class DeviceShareKey(
     @SerializedName("reconstructedETH") val reconstructedKey: String,
     @SerializedName("privateSOL") val privateKey: String,
     @SerializedName("deviceShare") val share: DeviceShare?,
+    @SerializedName("userId") var userId: String?,
 ) {
     data class DeviceShare(
```
gslevinkov, line 11: i would really want to special name entities that we serialize in shared prefs, what do you think?

app/src/main/res/layout/fragment_continue_onboarding.xml
```diff
@@ -0,0 +1,84 @@
...
+<androidx.constraintlayout.widget.ConstraintLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:id="@+id/contentView"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:background="@color/lime">
+
+    <LinearLayout
```
gslevinkov, line 11: fyi LinearLayout can be replaced with chains

app/src/main/java/org/p2p/wallet/auth/common/WalletWeb3AuthManager.kt
kabramovp2p: why not?
you want apply ?)

app/src/main/java/org/p2p/wallet/auth/common/WalletWeb3AuthManager.kt
kabramovp2p: oh yes I think you are right

gslevinkov: and do we need to manipulate raw string here? i think it's okay to return the whole entity from gson right here

app/src/main/res/layout/fragment_continue_onboarding.xml
kabramovp2p: is it works faster? as far as I remember complex chains worked not so good if they were more then one

app/src/main/java/org/p2p/wallet/auth/common/WalletWeb3AuthManager.kt
kabramovp2p: yeah we can do that

app/src/main/java/org/p2p/wallet/auth/model/DeviceShareKey.kt
kabramovp2p: what names ?
I do not get it

app/src/main/java/org/p2p/wallet/auth/common/WalletWeb3AuthManager.kt
gslevinkov: just sharedPreferences.contains :)

kuznetsov:
Screenshots in table would look much better

app/src/main/res/layout/fragment_continue_onboarding.xml
kuznetsov: I think it's just better to avoid layout nesting

kuznetsov (approved):

PR #616, opened 2022-08-02
PWN-4121 - use sol to pay compensations if no pools found

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-4121

## Description of Work

Use sol to pay compensations if no pools found

### DIFF
--- a/app/src/main/java/org/p2p/wallet/send/ui/main/SendPresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/send/ui/main/SendPresenter.kt
@@ -1,5 +1,8 @@
 package org.p2p.wallet.send.ui.main
 
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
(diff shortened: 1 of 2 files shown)


davran1001 (approved):

app/src/main/java/org/p2p/wallet/send/ui/main/SendPresenter.kt
```diff
@@ -705,6 +713,14 @@ class SendPresenter(
         return fees.feeInSol to fees.feeInPayingToken
     }
 
+    private fun showError() {
```
gslevinkov, line 716: can't agree with this naming
presenter can't show error
it can handle it, or request view to show error

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/send/ui/main/SendPresenter.kt
kabramovp2p: ok

PR #627, opened 2022-08-04
PWN-4393 - New main toolbar with balance block

kuznetsov:
### Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-4393
https://p2pvalidator.atlassian.net/browse/PWN-4394

#### Toolbar with balance block redesign
- Balance loading elements were added to xml but weren't implemented
- ActionButtons were temporary hidden

### Screenshots

Screenshot 2022-08-04 at 20 11 02


app/src/main/res/layout/fragment_home.xml
```diff
@@ -9,69 +9,35 @@
...
 
-            <LinearLayout
+            <include
+                android:id="@+id/balance"
+                layout="@layout/layout_home_balance"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
-                android:background="@color/backgroundPrimary"
-                android:orientation="vertical">
+                android:layout_marginTop="80dp"
```
kabramovp2p, line 32: it is better to add 80dp as toolbarSize
and use it here and in height of layout_home_toolbar

app/src/main/res/layout/fragment_home.xml
```diff
@@ -104,11 +70,12 @@
         </androidx.swiperefreshlayout.widget.SwipeRefreshLayout>
 
         <org.p2p.wallet.common.ui.widget.ActionButtonsView
-            android:id="@+id/actionButtonsView"
+            android:id="@+id/action_buttons_view"
```
kabramovp2p, line 73: naming in new style needed actionButtonsViewHome - for example

app/src/main/java/org/p2p/wallet/restore/interactor/SecretKeyInteractor.kt
```diff
@@ -88,7 +88,7 @@ class SecretKeyInteractor(
             putString(KEY_DERIVATION_PATH, path.stringValue)
         }
 
-        if (lookup) {
+        if (false) {
```
davran1001, line 91: Forgot to revert

davran1001 (approved):

app/src/main/java/org/p2p/wallet/restore/interactor/SecretKeyInteractor.kt
kuznetsov: Thanks!

kabramovp2p (approved):

PR #648, opened 2022-08-11
PWN-4521 - Toolbar profile click

kuznetsov:
### Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-4521

### Description of Work

Open `ReserveUsername` or `Settings` depending on username existence

### DIFF
--- a/app/src/main/java/org/p2p/wallet/home/ui/main/HomeContract.kt
+++ b/app/src/main/java/org/p2p/wallet/home/ui/main/HomeContract.kt
@@ -28,6 +28,6 @@ interface HomeContract {
         fun toggleTokenVisibility(token: Token.Active)
         fun toggleTokenVisibilityState()
         fun clearTokensCache()
-        fun hasUsername(): Boolean
+        fun usernameExists(): Boolean
     }
 }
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/p2p/wallet/home/ui/main/HomeContract.kt
```diff
@@ -28,5 +28,6 @@ interface HomeContract {
         fun toggleTokenVisibility(token: Token.Active)
         fun toggleTokenVisibilityState()
         fun clearTokensCache()
+        fun hasUsername(): Boolean
```
gslevinkov, line 31: im not sure that presenter can have a username :) 

i can only suggest Davran's idea - put it all in the presenter
Presenter will have only one method - onProfileIconClicked or smth like this

and presenter will have the logic and the if inside. 
```
if (username exists)
view.navigateToProfile
else
view.navigateToReserveName
```
making the view stupid as hell

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/home/ui/main/HomeContract.kt
kuznetsov: Nice point 🤘

PR #656, opened 2022-08-15
PWN-4447 New Pin Code fixes

kuznetsov:
### Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-4440
https://p2pvalidator.atlassian.net/browse/PWN-4446
https://p2pvalidator.atlassian.net/browse/PWN-4447
https://p2pvalidator.atlassian.net/browse/PWN-4464

### Description of Work

A bunch of fixes on new PinCode and Biometric screens

### DIFF
--- a/app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinContract.kt
+++ b/app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinContract.kt
@@ -14,11 +14,14 @@ interface NewCreatePinContract {
         fun vibrate(duration: Long)
         fun showLoading(isLoading: Boolean)
         fun navigateBack()
+        fun navigateToBiometrics(pinCode: String)
+        fun navigateToMain()
     }
 
     interface Presenter : MvpPresenter<View> {
         var pinMode: PinMode
         fun setPinCode(pinCode: String)
         fun onBackPressed()
+        fun onPinCreated()
     }
 }
--- a/app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinContract.kt
+++ b/app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinContract.kt
@@ -17,6 +17,8 @@ interface NewCreatePinContract {
     }
 
     interface Presenter : MvpPresenter<View> {
+        var pinMode: PinMode
         fun setPinCode(pinCode: String)
+        fun onBackPressed()
     }
 }
--- a/app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinFragment.kt
+++ b/app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinFragment.kt
@@ -33,14 +33,16 @@ class NewCreatePinFragment :
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
+        presenter.pinMode = PinMode.CREATE
+
         with(binding) {
-            toolbar.setNavigationOnClickListener { popBackStack() }
+            toolbar.setNavigationOnClickListener { presenter.onBackPressed() }
             pinView.onPinCompleted = { presenter.setPinCode(it) }
             pinView.onKeyboardClicked = { vibrate(VIBRATE_DURATION) }
         }
 
         requireActivity().onBackPressedDispatcher.addCallback(viewLifecycleOwner) {
-            popBackStack()
+            presenter.onBackPressed()
         }
     }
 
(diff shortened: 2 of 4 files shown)


kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinFragment.kt
```diff
@@ -33,14 +35,16 @@ class NewCreatePinFragment :
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
+        presenter.pinMode = PinMode.CREATE
```
gslevinkov, line 38: no need to put explicitly put this value here, it has default in Presenter

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinContract.kt
```diff
@@ -14,9 +14,14 @@ interface NewCreatePinContract {
         fun vibrate(duration: Long)
         fun showLoading(isLoading: Boolean)
         fun navigateBack()
+        fun navigateToBiometrics(pinCode: String)
+        fun navigateToMain()
     }
 
     interface Presenter : MvpPresenter<View> {
+        var pinMode: PinMode
```
1sakov, line 22: Why don't you just create a global value inside presenter and leave here a set method? 
Like `setMode` or smth.
We don't have variables in presenters

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinFragment.kt
kuznetsov: Actually is has to reset to default mode in case of returning back from next screen

app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinContract.kt
kuznetsov: Property style is better as for me, but let's make it consistent, okay

app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinPresenter.kt
```diff
@@ -1,20 +1,37 @@
...
 import org.p2p.wallet.common.mvp.BasePresenter
 import org.p2p.wallet.utils.emptyString
+import timber.log.Timber
 
 private const val VIBRATE_DURATION = 500L
 
+enum class PinMode {
+    CREATE,
+    CONFIRM
+}
```
1sakov, line 19: Are you sure it's the right place for that class? 
I would move as a separate file

app/src/main/java/org/p2p/wallet/splash/SplashFragment.kt
```diff
@@ -18,7 +18,8 @@ class SplashFragment :
     override val presenter: SplashContract.Presenter by inject()
 
     override fun navigateToOnboarding() {
-        replaceFragment(OnboardingRootFragment.create())
+        // TODO: Replace back with OnboardingRootFragment after QA
+        replaceFragment(NewCreatePinFragment.create())
```
1sakov, line 22: Don't forget to do that

ui-kit/src/main/res/layout/widget_ui_kit_pin_code_view_dot.xml
```diff
@@ -4,6 +4,6 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="12dp"
     android:layout_height="12dp"
-    android:layout_margin="12dp"
+    android:layout_margin="6dp"
```
1sakov, line 7: Are you sure it's not 8 or 4?

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinPresenter.kt
kuznetsov: Ok, let's separate it

ui-kit/src/main/res/layout/widget_ui_kit_pin_code_view_dot.xml
kuznetsov: It's 12dp between dots, so that's right

1sakov (approved):

PR #663, opened 2022-08-16
PWN-4267 - onboarding fixes

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-4267

## Description of Work

PWN-4434 - make toast according to designs
PWN-4428 - keep calm if there is no connection

### DIFF
--- a/app/src/main/java/org/p2p/wallet/auth/web3authsdk/GoogleSignInHelper.kt
+++ b/app/src/main/java/org/p2p/wallet/auth/web3authsdk/GoogleSignInHelper.kt
@@ -12,15 +12,27 @@ import com.google.android.gms.auth.api.identity.SignInClient
 import com.google.android.gms.auth.api.identity.SignInCredential
 import com.google.android.gms.common.api.ApiException
 import org.p2p.wallet.R
+import org.p2p.wallet.common.ResourcesProvider
+import org.p2p.wallet.common.mvp.MvpView
+import org.p2p.wallet.updates.ConnectionStateProvider
 import timber.log.Timber
 import java.util.UUID
 
 private const val USER_CANCELED_DIALOG_CODE = 16
 
-class GoogleSignInHelper() {
+class GoogleSignInHelper(
+    private val connectionStateProvider: ConnectionStateProvider,
+    private val resourcesProvider: ResourcesProvider
+) {
+
+    private var view: MvpView? = null
 
     private fun getSignInClient(context: Context): SignInClient {
         return Identity.getSignInClient(context)
+    }
+
+    fun attach(mvpView: MvpView) {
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/p2p/wallet/auth/web3authsdk/GoogleSignInHelper.kt
```diff
@@ -52,10 +64,18 @@ class GoogleSignInHelper() {
             getSignInClient(context).getSignInCredentialFromIntent(result.data)
         } catch (ex: ApiException) {
             if (ex.statusCode != USER_CANCELED_DIALOG_CODE) {
-                Toast.makeText(context, ex.toString(), Toast.LENGTH_SHORT).show()
+                if (connectionStateProvider.hasConnection()) {
+                    view?.showErrorSnackBar(ex.message ?: ex.toString())
```
gslevinkov, line 68: i would rather pass callback or idk, instead of working with view
Helper should be stupid i think

app/src/main/java/org/p2p/wallet/common/mvp/BaseMvpFragment.kt
```diff
@@ -78,4 +78,14 @@ abstract class BaseMvpFragment<V : MvpView, P : MvpPresenter<V>>(
                 .setAction(actionResId, block)
         }
     }
+
+    override fun showInfoSnackBar(message: String, iconResId: Int?, actionResId: Int?, block: (() -> Unit)?) {
```
gslevinkov, line 82: block -> actionBlock

app/src/main/java/org/p2p/wallet/auth/web3authsdk/GoogleSignInHelper.kt
kabramovp2p: i can make it in a callback
i do not mind

1sakov (approved):

app/src/main/java/org/p2p/wallet/auth/web3authsdk/GoogleSignInHelper.kt
gslevinkov: well, i meant pass callback as a parameter really :) 
It would leave us with only one method needed instead of three: attach / method that we need / detach.
this all detach/attach thing is not needed here - listener / handler is not Context-aware thing, there will be no leak.

kabramovp2p: Fuuuuuu! memory leak

gslevinkov (approved):

PR #677, opened 2022-08-22
PWN-4690 - check error on valid exception and do not show it if not needed

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-4690

## Description of Work

Check error on valid exception and do not show it if not needed


app/src/main/java/org/p2p/wallet/auth/web3authsdk/GoogleSignInHelper.kt
```diff
@@ -70,6 +71,11 @@ class GoogleSignInHelper(
         }
     }
 
+    private fun isValidError(exception: ApiException): Boolean {
```
gslevinkov, line 74: isValidError -> shouldErrorBeHandled
that's what method intents to do, it checks should we handle error from google

gslevinkov (approved):

1sakov (approved):

PR #678, opened 2022-08-23
PWN-4507 - Phone mask and hint corrected

1sakov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-4507

## Description of Work

Removed hint draw in onDraw method
Created `HintTextView` for showing hint
Showing number hints instead of skin

## Screenshots (Optional)

Screenshot_20220823_160720 | Screenshot_20220823_160751 | Screenshot_20220823_160801

### DIFF
--- a/app/src/main/java/org/p2p/wallet/auth/widget/HintEditText.kt
+++ b/app/src/main/java/org/p2p/wallet/auth/widget/HintEditText.kt
@@ -1,45 +1,19 @@
-/*
- * This is the source code of Telegram for Android v. 5.x.x.
- * It is licensed under GNU GPL v. 2 or later.
- * You should have received a copy of the license in this archive (see LICENSE).
- *
- * Copyright Nikolai Kudashov, 2013-2018.
- */
 package org.p2p.wallet.auth.widget
 
 import androidx.appcompat.widget.AppCompatEditText
 import android.content.Context
-import android.content.res.Resources
-import android.graphics.Canvas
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/p2p/wallet/auth/widget/HintEditText.kt
```diff
@@ -1,85 +1,29 @@
...
-    constructor(context: Context?, attrs: AttributeSet?) : super(context!!, attrs) {
-        paint.color = Color.LTGRAY
-    }
-
-    constructor(context: Context?, attrs: AttributeSet?, defStyleAttrs: Int) : super(
-        context!!, attrs, defStyleAttrs
-    ) {
-        paint.color = Color.LTGRAY
-    }
+    constructor(context: Context) : super(context)
```
gslevinkov, line 14: better with @JvmOverloads as we always do

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/auth/widget/HintEditText.kt
1sakov: It's not working, the text style is being dropped. 
I didn't have time to debug and decided to revert back as it now.

But first, I made as we always do.

Let's skip this one for now, have no time for investigation

kabramovp2p (approved):

PR #684, opened 2022-08-25
PWN-4730 - do not call API on same number

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-4730 

## Description of Work

Save last phone number what user make to auth - and do not call api if it is same


app/src/main/java/org/p2p/wallet/auth/ui/phone/PhoneNumberEnterPresenter.kt
```diff
@@ -68,7 +70,11 @@ class PhoneNumberEnterPresenter(
         launch {
             try {
                 selectedCountryCode?.let {
-                    createWalletInteractor.startCreatingWallet(userPhoneNumber = it.phoneCode + phoneNumber)
+                    val userPhoneNumber = it.phoneCode + phoneNumber
+                    if (lastPhoneNumber != userPhoneNumber) {
+                        createWalletInteractor.startCreatingWallet(userPhoneNumber = it.phoneCode + phoneNumber)
```
gslevinkov, line 75:
```suggestion
                        createWalletInteractor.startCreatingWallet(userPhoneNumber = userPhoneNumber)
```

gslevinkov (approved):

1sakov (approved):

PR #696, opened 2022-08-30
PWN-4464 fix bug after pin create

davran1001:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-4464

## Description of Work

Fix navigation to main screen after creating PIN code

### DIFF
--- a/app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinPresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinPresenter.kt
@@ -1,6 +1,5 @@
 package org.p2p.wallet.auth.ui.pin.newcreate
 
-import org.p2p.wallet.R
 import org.p2p.wallet.auth.analytics.AdminAnalytics
 import org.p2p.wallet.auth.interactor.AuthInteractor
 import org.p2p.wallet.auth.interactor.AuthLogoutInteractor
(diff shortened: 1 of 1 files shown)


gslevinkov:
no assignee
wrong PR name

app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinPresenter.kt
```diff
@@ -70,7 +70,6 @@ class NewCreatePinPresenter(
 
     private fun checkBiometricAvailability() {
         if (authInteractor.getBiometricStatus() < BiometricStatus.AVAILABLE) {
-            createWalletInteractor.finishAuthFlow()
```
gslevinkov, line 73: @kuznetsov fyi

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinPresenter.kt
kuznetsov: It's not okay :(

kuznetsov: We have to finish auth  flow if user has no biometrics and pin is enough to start main screen. But now looks like instead of mock web3auth we can't generate user and finishAuthFlow() is crashing

davran1001: i revert this

kabramovp2p (approved):

PR #698, opened 2022-08-30
PWN-4440 - Add toast on back button pressed

davran1001:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-4449

## Description of Work

Add toast message, when user click on back button, in CreatePinFragment, also close application if it was clicked twice

### DIFF
--- a/app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinPresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinPresenter.kt
@@ -1,5 +1,6 @@
 package org.p2p.wallet.auth.ui.pin.newcreate
 
+import org.p2p.wallet.R
 import org.p2p.wallet.auth.analytics.AdminAnalytics
 import org.p2p.wallet.auth.interactor.AuthInteractor
 import org.p2p.wallet.auth.interactor.AuthLogoutInteractor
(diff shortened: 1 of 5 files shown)


kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinPresenter.kt
```diff
@@ -22,6 +21,7 @@ class NewCreatePinPresenter(
 
     private var createdPin = emptyString()
     private var pinMode = PinMode.CREATE
+    private var isOnBackButtonPressed = false
```
gslevinkov, line 24: isOnBackButtonPressed  -> navigateBackOnBackPressed
we need to put some logic behind this name that is connected with exiting app when back is pressed twice

app/src/main/java/org/p2p/wallet/splash/SplashFragment.kt
```diff
@@ -1,5 +1,6 @@
 package org.p2p.wallet.splash
 
+import android.view.View
```
gslevinkov, line 3: ?

gslevinkov (approved):

kabramovp2p (approved):

kuznetsov (approved):

PR #703, opened 2022-09-01
PWN-4784 - Country code reverted back sorting

1sakov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-4784

## Description of Work

## Screenshots (Optional)

### DIFF
--- a/app/src/main/java/org/p2p/wallet/auth/ui/phone/countrypicker/CountryCodePickerPresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/auth/ui/phone/countrypicker/CountryCodePickerPresenter.kt
@@ -16,6 +16,7 @@ class CountryCodePickerPresenter(
     BasePresenter<CountryCodePickerContract.View>(),
     CountryCodePickerContract.Presenter {
 
+    private var selectedCountryCode: CountryCode? = null
     private var searchText: String = emptyString()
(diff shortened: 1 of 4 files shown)


davran1001 (approved):

app/src/main/java/org/p2p/wallet/auth/ui/phone/countrypicker/CountryCodePickerPresenter.kt
```diff
@@ -16,6 +16,7 @@ class CountryCodePickerPresenter(
     BasePresenter<CountryCodePickerContract.View>(),
     CountryCodePickerContract.Presenter {
 
+    private var selectedCountryCode: CountryCode? = null
```
gslevinkov, line 19: don't we have defaultCountryCode that is selected on screen start?

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/auth/ui/phone/countrypicker/CountryCodePickerPresenter.kt
1sakov: looks like no, but i'm not sure. 
I just reverted back changes.

Testing is okay, so looks like it's ok

PR #706, opened 2022-09-05
PWN-4716 - Change PIN code flow

1sakov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-4716

## Description of Work

Change pin code flow from settings added 

## Screenshots (Optional)

Screenshot_20220905_132003 | Screenshot_20220905_131753 | Screenshot_20220905_131801


app/src/main/java/org/p2p/wallet/settings/ui/newreset/pin/NewForgotPinBottomSheet.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.p2p.wallet.common.ui.NonDraggableBottomSheetDialogFragment
+import org.p2p.wallet.databinding.DialogNewForgotPasswordBinding
+import org.p2p.wallet.utils.args
+import org.p2p.wallet.utils.viewbinding.viewBinding
+import org.p2p.wallet.utils.withArgs
+
+private const val EXTRA_REQUEST_KEY = "EXTRA_REQUEST_KEY"
+private const val EXTRA_RESULT_KEY = "EXTRA_RESULT_KEY"
+
+class NewForgotPinBottomSheet : NonDraggableBottomSheetDialogFragment() {
```
gslevinkov, line 20: it's better to add NonDraggableBottomSheetDialogFragment to proguard then

app/src/main/java/org/p2p/wallet/settings/ui/newreset/pin/NewForgotPinBottomSheet.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    companion object {
+        fun show(fm: FragmentManager, requestKey: String, resultKey: String) =
+            NewForgotPinBottomSheet().withArgs(
+                EXTRA_REQUEST_KEY to requestKey,
+                EXTRA_RESULT_KEY to resultKey
+            ).show(fm, NewForgotPinBottomSheet::javaClass.name)
```
gslevinkov, line 27:
```suggestion
            NewForgotPinBottomSheet()
                .withArgs(
                    EXTRA_REQUEST_KEY to requestKey,
                    EXTRA_RESULT_KEY to resultKey
                )
                .show(fm, NewForgotPinBottomSheet::javaClass.name)
```
every line break - a new sentence, a new intent like in Russian language

you create fragment
you add some args
you show it

not 
you create fragment add some args
       you show it

app/src/main/java/org/p2p/wallet/settings/ui/newreset/pin/NewForgotPinBottomSheet.kt
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    private val binding: DialogNewForgotPasswordBinding by viewBinding()
+    private val requestKey: String by args(EXTRA_REQUEST_KEY)
+    private val resultKey: String by args(EXTRA_RESULT_KEY)
+
+    override fun getTheme(): Int = R.style.WalletTheme_BottomSheet_Rounded
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View =
+        inflater.inflate(R.layout.dialog_new_forgot_password, container, false)
```
gslevinkov, line 37: do we really want to use inflater all the time? how about a tech task that would create a base class that would help us to remove onCreateView on bottom sheets?

app/src/main/java/org/p2p/wallet/settings/ui/newreset/pin/NewForgotPinBottomSheet.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        with(binding) {
+            buttonLogout.setOnClickListener {
+                setFragmentResult(requestKey, bundleOf(resultKey to true))
```
gslevinkov, line 43: im not sure what is `true` here
we send true when button is clicked? or user confirmed logout? or he is confirming that he forgot password / pin?

maybe we should rename resultKey to provide some intent what value he holds inside? 
smth like resultKeyLogoutConfirmed
this bottom sheet is not universal, we can add some context to request and result key naming

davran1001 (approved):

app/src/main/java/org/p2p/wallet/settings/ui/newreset/pin/NewResetPinFragment.kt
```diff
@@ -0,0 +1,113 @@
...
+                )
+            }
+        }
+    }
+
+    override fun navigateToOnboarding() {
+        popAndReplaceFragment(OnboardingRootFragment.create(), inclusive = true)
+    }
+
+    override fun showResetSuccess() {
```
gslevinkov, line 79: well, we don't really show some stuff on the screen
how about finishResetPassword / smth with navigate word

app/src/main/java/org/p2p/wallet/settings/ui/newreset/pin/NewResetPinPresenter.kt
```diff
@@ -0,0 +1,103 @@
...
+
+    private fun resetPinActually() {
+        launch {
+            try {
+                authInteractor.resetPin(createdPin)
+                view?.showMessage(R.string.settings_item_pin_changed)
+                view?.showResetSuccess()
+            } catch (e: Throwable) {
+                Timber.e(e, "error setting new pin")
+                view?.showErrorMessage(e)
```
gslevinkov, line 95: !!! showErrorMessage would use old bottom sheet instead of ui kit snackbar, please change

app/src/main/java/org/p2p/wallet/settings/ui/newreset/pin/NewResetPinPresenter.kt
```diff
@@ -0,0 +1,103 @@
...
+                view?.showPinCorrect()
+                view?.vibrate(VIBRATE_DURATION)
+            } finally {
+                createdPin = emptyString()
```
gslevinkov, line 99: nit: do we want to manipulate strings here instead of CharArray? CharArray won't hold the pin code in the String pool and it's safer

app/src/main/java/org/p2p/wallet/settings/ui/settings/presenter/SettingsItem.kt
```diff
@@ -25,7 +25,7 @@ sealed interface SettingsItem {
     ) : SettingsItem
 
     data class ComplexSettingsItem(
-        @StringRes val settingNameRes: Int,
+        @StringRes val nameRes: Int,
```
gslevinkov, line 28: why not settingsName? don't you think that name here is too abstract? is it item name or user name? i would rather be more specific here, it's always not redundant

app/src/main/res/drawable/ic_dot.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="oval">
+
+    <size
+        android:width="4dp"
+        android:height="4dp" />
+    <solid android:color="@color/night" />
```
gslevinkov, line 9: icons_night

app/src/main/res/drawable/bg_snow_top_rounded_medium.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle">
+
+    <corners
+        android:topLeftRadius="16dp"
+        android:topRightRadius="16dp" />
+
+    <solid android:color="@color/snow" />
```
gslevinkov, line 10: icons_snow

app/src/main/res/layout/dialog_new_forgot_password.xml
```diff
@@ -0,0 +1,45 @@
...
+        android:id="@+id/buttonLogout"
+        style="@style/UiKit.Components.Button.Large"
+        android:layout_width="match_parent"
+        android:layout_height="@dimen/ui_kit_button_large_height"
+        android:layout_marginHorizontal="24dp"
+        android:layout_marginTop="32dp"
+        android:layout_marginBottom="16dp"
+        android:backgroundTint="@color/button_rain"
+        android:text="@string/auth_logout"
+        android:textColor="@color/night" />
```
gslevinkov, line 44: text_night

gslevinkov (approved):

kabramovp2p (approved):
feel free to merge after Gleb's comments fixes

app/src/main/java/org/p2p/wallet/settings/ui/newreset/pin/NewForgotPinBottomSheet.kt
1sakov: yeah, agree. 
just copy pasted

app/src/main/java/org/p2p/wallet/settings/ui/newreset/pin/NewForgotPinBottomSheet.kt
1sakov: Why?
Why should we add bottom sheet to proguard?

app/src/main/java/org/p2p/wallet/settings/ui/newreset/pin/NewForgotPinBottomSheet.kt
1sakov: not in this PR.

app/src/main/java/org/p2p/wallet/settings/ui/newreset/pin/NewForgotPinBottomSheet.kt
1sakov: We are not using the value, I just need to send some bundle to check if user clicked on logout button

app/src/main/java/org/p2p/wallet/settings/ui/newreset/pin/NewResetPinFragment.kt
1sakov: sounds good

app/src/main/java/org/p2p/wallet/settings/ui/newreset/pin/NewForgotPinBottomSheet.kt
gslevinkov: i saw in sentry that proguard minifies fragment names for bottom sheets, gotta check it out, maybe i already added this rule

app/src/main/java/org/p2p/wallet/settings/ui/newreset/pin/NewForgotPinBottomSheet.kt
gslevinkov: yeah yeah, i wrote about a tech task, not to change smth here

app/src/main/java/org/p2p/wallet/settings/ui/newreset/pin/NewForgotPinBottomSheet.kt
gslevinkov: how about a Unit?

app/src/main/java/org/p2p/wallet/settings/ui/newreset/pin/NewResetPinPresenter.kt
1sakov: I reused the implementation which works perfect since it was developed. 
We can consider moving top CharArray as a tech debt. 

Refactoring it now may cause additional bugs, we don't have time

app/src/main/java/org/p2p/wallet/settings/ui/settings/presenter/SettingsItem.kt
1sakov: Because we already have the class name.
Settings item has settings id seems redundant for sure.
And it's the name of the settings, not the settings name

app/src/main/java/org/p2p/wallet/settings/ui/newreset/pin/NewForgotPinBottomSheet.kt
1sakov: I think it's expected behavior. 
We don't minify the class if it breaks the logic only. 
In other cases, we should be good

app/src/main/java/org/p2p/wallet/settings/ui/newreset/pin/NewForgotPinBottomSheet.kt
1sakov: > how about a Unit?

first, I passed Unit, but it crashed. Looks like we cannot pass Unit in bundle

PR #707, opened 2022-09-05
PWN-4895 - Profile screen bugfixes

1sakov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-4896
https://p2pvalidator.atlassian.net/browse/PWN-4895
https://p2pvalidator.atlassian.net/browse/PWN-4894
https://p2pvalidator.atlassian.net/browse/PWN-4893

## Description of Work

Various bugfixes made

## Screenshots (Optional)


gslevinkov:
i would suggest to add task numbers to the commit body to link them in Jira

app/src/main/java/org/p2p/wallet/settings/ui/settings/NewSettingsPresenter.kt
```diff
@@ -60,15 +60,14 @@ class NewSettingsPresenter(
...
-        loadSettings()
     }
 
     override fun onBiometricSignInSwitchChanged(isSwitched: Boolean) {
         if (isSwitched) {
             view?.confirmBiometrics(authInteractor.getPinEncodeCipher())
         } else {
             authInteractor.disableBiometricSignIn()
-            loadSettings()
+            view?.updateSwitchItem(R.string.settings_item_title_touch_id, isSwitched = false)
```
gslevinkov, line 70: load settings should get the current value of the biometric and update all the switches accordingly 
it doesn't work as expected?

app/src/main/java/org/p2p/wallet/settings/ui/settings/adapter/NewSettingsAdapter.kt
```diff
@@ -1,62 +1,94 @@
...
-        items.addAll(newItems)
-        notifyDataSetChanged()
+        val old = data.toMutableList()
+        data.clear()
+        data.addAll(newItems)
+        DiffUtil.calculateDiff(SettingsItemsDiffCallback(old, data)).dispatchUpdatesTo(this)
+    }
+
+    fun updateSwitchItem(itemId: Int, isSwitched: Boolean) {
+        val itemIndex = data.indexOfFirst { it is SwitchSettingsItem && it.nameRes == itemId }
```
gslevinkov, line 38: im not sure that updating switch item manually is a good idea
we should always fetch the latest actual data from the storage and update the screen from it, so it's one side data flow

1sakov: This method is being updated from the latest storage. 
And this is exceptional case, since the checkbox updates itself. 
And we are updating its state from presenter additionally. 
That's why we need it. 

In general, for sure we should always have one way data flow

PR #769, opened 2022-09-20
PWN-5175 - on lockout added and connected to SignInPinFragment

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5175

## Description of Work

OnLockout added and connected to SignInPinFragment

### DIFF
--- a/app/src/main/java/org/p2p/wallet/auth/ui/pin/signin/SignInPinFragment.kt
+++ b/app/src/main/java/org/p2p/wallet/auth/ui/pin/signin/SignInPinFragment.kt
@@ -3,6 +3,7 @@ package org.p2p.wallet.auth.ui.pin.signin
 import android.os.Bundle
 import android.view.View
 import androidx.activity.addCallback
+import androidx.annotation.StringRes
 import androidx.appcompat.app.AlertDialog
 import org.koin.android.ext.android.inject
 import org.p2p.wallet.R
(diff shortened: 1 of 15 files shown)


app/src/main/java/org/p2p/wallet/auth/ui/pin/signin/SignInPinFragment.kt
```diff
@@ -35,7 +36,17 @@ class SignInPinFragment :
     private val binding: FragmentSignInPinBinding by viewBinding()
     private val analyticsInteractor: ScreensAnalyticsInteractor by inject()
     private val biometricWrapper by lazy {
-        BiometricPromptWrapper(this) { presenter.signInByBiometric(it) }
+        BiometricPromptWrapper(
+            this,
+            onLockout = {
```
gslevinkov, line 41: we do not use parameters passed in `onLockout`

app/src/main/java/org/p2p/wallet/utils/BiometricPromptWrapper.kt
```diff
@@ -14,19 +14,24 @@ class BiometricPromptWrapper(
...
 
         override fun onAuthenticationError(errorCode: Int, errString: CharSequence) {
-            onError?.invoke(
-                errString.takeIf {
-                    errorCode != BiometricPrompt.ERROR_USER_CANCELED &&
-                        errorCode != BiometricPrompt.ERROR_NEGATIVE_BUTTON
-                }
-                    ?.toString()
-            )
+            if (errorCode == BiometricPrompt.ERROR_LOCKOUT) {
```
gslevinkov, line 24: are we logging stuff here? smth like Timber.i?

gslevinkov (changes requested):

app/src/main/java/org/p2p/wallet/utils/BiometricPromptWrapper.kt
kabramovp2p: no logs
why do we need this?

gslevinkov (approved):

PR #770, opened 2022-09-20
PWN-5209 - ability to delete device share from debug meny

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5209

## Description of Work

Show device share and have an ability to remove it from Debug settings

### DIFF
--- a/app/src/main/java/org/p2p/wallet/debug/torus/DebugTorusFragment.kt
+++ b/app/src/main/java/org/p2p/wallet/debug/torus/DebugTorusFragment.kt
@@ -66,7 +66,9 @@ class DebugTorusFragment : BaseFragment(R.layout.fragment_debug_torus) {
             val hasShare = signUpDetails?.signUpDetails?.deviceShare != null
             shareTextView.text = if (hasShare) {
                 signUpDetails?.signUpDetails?.deviceShare.toString()
-            } else "No device share"
+            } else {
+                "No device share"
+            }
             shareDeleteButton.apply {
                 isEnabled = hasShare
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/p2p/wallet/debug/torus/DebugTorusFragment.kt
```diff
@@ -58,6 +61,19 @@ class DebugTorusFragment : BaseFragment(R.layout.fragment_debug_torus) {
                 val newVerifier = verifierEditText.text.toString()
                 updateEnvironmentAndRestart(newUrl = newUrl, newVerifier = newVerifier)
             }
+
+            val signUpDetails = signUpDetailsStorage.getLastSignUpUserDetails()
+            val hasShare = signUpDetails?.signUpDetails?.deviceShare != null
+            shareTextView.text = if (hasShare) {
+                signUpDetails?.signUpDetails?.deviceShare.toString()
+            } else "No device share"
```
gslevinkov, line 69:
```suggestion
            shareTextView.text = if (hasShare) {
                signUpDetails?.signUpDetails?.deviceShare.toString()
            } else { 
              "No device share"
            }
```

also - deviceShare.toString() will represent the whole object in String representation. do you need a string share value here or a whole object with verifiers and stuff?

kabramovp2p: i think it will be cool to check some data if QA needs

PR #777, opened 2022-09-21
PWN-5182 - set starting screen except of seed restore

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5182

## Description of Work

Set starting screen button except of seed restore button


app/src/main/java/org/p2p/wallet/auth/ui/restore/common/CommonRestoreFragment.kt
```diff
@@ -129,6 +130,16 @@ class CommonRestoreFragment :
                     text = userId
                 }
                 textViewTryAnother.isVisible = true
+
+                buttonSeed.apply {
```
gslevinkov, line 134: rename the view id then

gslevinkov (approved):

davran1001 (approved):

PR #789, opened 2022-09-23
PWN-5258 - generate mnemonics on client and pass it to Web3Auth

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5258


app/src/main/java/org/p2p/wallet/auth/web3authsdk/Web3AuthApiClient.kt
```diff
@@ -136,14 +141,21 @@ class Web3AuthApiClient(
         val torusEndpoint = "$host:5051"
         val torusVerifier = torusNetwork.verifier
 
+        if (userGeneratedSeedPhrase.isEmpty()) {
```
kabramovp2p, line 144: we need to clarify when we need to add this privateInput with @turisap 
because he told that we need it only in SignUp methods

but if on signIn it will be ignored i think it is also ok

kabramovp2p (approved):

PR #795, opened 2022-09-26
PWN-5213 - update Client API Gateway with encrypted_metadata

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5213

## Description of Work

- derive new secret key from new derivation path
- encrypt json with ChaCha algorithm


app/src/main/java/org/p2p/wallet/auth/gateway/repository/mapper/GatewayServiceCreateWalletMapper.kt
```diff
@@ -101,36 +111,45 @@ class GatewayServiceCreateWalletMapper(
...
+            )
+        )
 
         val signatureField: Base58String = signatureFieldGenerator.generateSignatureField(
             userPrivateKey = userPrivateKey,
             structToSerialize = ConfirmRegisterWalletSignatureStruct(
                 clientId = userPublicKey.base58Value,
                 etheriumId = etheriumAddress.lowercase(),
-                phone = phoneNumber,
-                encryptedShare = gson.toJson(thirdShare),
```
gslevinkov, line 118: "not a valid signature" returned from the gateway due to `gson.toJson` formatted JSON output

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/auth/gateway/repository/mapper/ChaCha20Poly1305Encryptor.kt
```diff
@@ -0,0 +1,37 @@
...
+
+        val finalOff = chaCha20Poly1305.processBytesKt(
+            inBytes = dataToEncrypt,
+            len = dataToEncrypt.size,
+            outBytes = encryptionResult,
+        )
+        chaCha20Poly1305.doFinal(encryptionResult, finalOff)
+
+        chaCha20Poly1305.reset()
+        return encryptionResult
```
1sakov, line 35: @gslevinkov can you ask somebody from management if it's okay that we show what type of encryption we do here?
Since it's an open-source, it may be dangerous, but I am not sure

1sakov (approved):

davran1001 (approved):

PR #801, opened 2022-09-27
PWN-5301 - SolendSDK: so files and JNI interface

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5301

## Description of Work

- .so files and JNI interface

### DIFF
--- a/app/src/main/java/P2pSdk.kt
+++ b/app/src/main/java/P2pSdk.kt
@@ -15,6 +15,12 @@ typealias u32 = UInt
 
 @Keep
 class P2pSdk {
+    companion object {
+        init {
+            System.loadLibrary("p2p_sdk_android")
+        }
+    }
+
     // for testing
     external fun greet(
         name: String
--- a/app/src/main/java/P2pSdk.kt
+++ b/app/src/main/java/P2pSdk.kt
@@ -1,0 +1,84 @@
+import androidx.annotation.Keep
+
+typealias jboolean = Boolean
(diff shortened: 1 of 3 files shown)


app/src/main/java/P2pSdk.kt
```diff
@@ -0,0 +1,84 @@
...
+typealias JInt = Int
+typealias jlong = Long
+typealias JFloat = Float
+typealias JDouble = Double
+typealias JString = String
+
+typealias u64 = ULong
+typealias u32 = UInt
+
+@Keep
```
gslevinkov, line 16: keep for proguard

app/src/main/java/org/p2p/wallet/App.kt
```diff
@@ -45,8 +44,14 @@ import org.p2p.wallet.user.UserModule
 import org.p2p.wallet.user.repository.prices.di.TokenPricesModule
 import org.p2p.wallet.utils.SolanajTimberLogger
 import timber.log.Timber
+import kotlinx.coroutines.launch
 
 class App : Application() {
+    companion object {
+        init {
```
kabramovp2p, line 51: maybe make this call in class P2pSdk  ? not here?
then logic will be located in one class with initialisation?

kabramovp2p (approved):

1sakov (approved):
feel free to merge after Kostya's comments

davran1001 (approved):

PR #802, opened 2022-09-27
PWN-3713 - add subVerifier and changes to support 1.2.0 Web3Auth

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3713

## Description of Work

PWN-3713 - add subVerifier and changes to support 1.2.0 Web3Auth

### DIFF
--- a/app/src/main/java/org/p2p/wallet/debug/torus/DebugTorusFragment.kt
+++ b/app/src/main/java/org/p2p/wallet/debug/torus/DebugTorusFragment.kt
@@ -33,11 +33,7 @@ class DebugTorusFragment : BaseFragment(R.layout.fragment_debug_torus) {
 
             networkServicesUrlProvider.loadTorusEnvironment().apply {
                 environmentTextView.text = baseUrl
-                verifierTextView.text = if (subVerifier.isNullOrBlank()) {
-                    verifier
-                } else {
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/p2p/wallet/debug/torus/DebugTorusFragment.kt
```diff
@@ -33,7 +33,11 @@ class DebugTorusFragment : BaseFragment(R.layout.fragment_debug_torus) {
 
             networkServicesUrlProvider.loadTorusEnvironment().apply {
                 environmentTextView.text = baseUrl
-                verifierTextView.text = verifier
+                verifierTextView.text = if (subVerifier.isNullOrBlank()) {
+                    verifier
+                } else {
+                    "verifier = $verifier\nsubVerifier = $subVerifier"
+                }
```
gslevinkov, line 40: it's always to put explicitly that there is subVerifier here instead of making it gone
```suggestion
                verifierTextView.text = "verifier = $verifier\nsubVerifier = ${subVerifier ?: "-"}"
```
nothing wrong with view looking like this
verifier = smth
subVerifier = -

gslevinkov (approved):

PR #803, opened 2022-09-27
PWN-5302 - SolendSdk. Mapping and errors

gslevinkov:
before checking this PR, check #801 
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5302

## Description of Work

- mapping and errors


app/src/main/java/solendsdkfacade/SolendMethodResultHandler.kt
```diff
@@ -0,0 +1,9 @@
+package solendsdkfacade
+
+import solendsdkfacade.model.SolendMethodResultError
+import solendsdkfacade.model.SolendMethodResultSuccess
+
+interface SolendMethodResultHandler {
```
gslevinkov, line 6: Handler in future - is a foreground service / workmanager that handles async result from SDK and works with UI to show toasts and stuff

app/src/main/java/solendsdkfacade/SolendSdkFacade.kt
```diff
@@ -0,0 +1,104 @@
...
+        currentBlockhash: String,
+        remainingFreeTransactionsCount: UInt,
+        payFeeWithRelay: Boolean,
+        feePayerToken: SolendFeePayerTokenData?,
+        realFeePayerAddress: Base58String
+    ) {
+        withContext(dispatchers.io) {
+            Timber.tag(TAG).i("Method createSolendDepositTransactions called")
+
+            val result = p2pSdk.createSolendDepositTransactions(
```
gslevinkov, line 48: result from SolendSDK is always string JSON

app/src/main/java/solendsdkfacade/mapper/SolendMethodResultMapper.kt
```diff
@@ -0,0 +1,31 @@
...
+import timber.log.Timber
+
+class SolendMethodResultMapper(
+    // public due to inline
+    val gson: Gson
+) {
+
+    inline fun <reified SuccessType : SolendMethodResultSuccess> fromSdk(sdkResult: String): SolendResult<SuccessType> {
+        return SolendResult(
+            success = gson.fromJsonOrNull(sdkResult),
```
gslevinkov, line 16: works like`kotlin.Result` class
try parsing the string as success class and error class

app/src/main/java/solendsdkfacade/model/SolendEnvironment.kt
```diff
@@ -0,0 +1,5 @@
+package solendsdkfacade.model
+
+enum class SolendEnvironment(val sdkValue: String) {
```
gslevinkov, line 3: in future - we can switch SolendEnvironment depending on BuildConfig or Debug panel

app/src/main/java/solendsdkfacade/model/SolendFeePayerTokenData.kt
```diff
@@ -0,0 +1,19 @@
...
+    @SerializedName("sender_account_pubkey")
+    val senderAccountAddress: Base58String,
+    @SerializedName("recipient_account_pubkey")
+    val recipientAccountAddress: Base58String,
+    @SerializedName("mint_pubkey")
+    val tokenMintAddress: Base58String,
+    @SerializedName("authority_pubkey")
+    val transferAuthorityAddress: Base58String,
+    @SerializedName("exchange_rate")
+    val tokenExchangeRateLamports: Double,
```
gslevinkov, line 16: question: lamports is Double type?

1sakov: no, lamports == BigInteger

app/src/main/java/org/p2p/wallet/AppModule.kt
```diff
@@ -30,5 +54,37 @@ object AppModule {
                 tokenKeyProvider = get()
             )
         }
+
+        includes(
```
1sakov, line 58: I am not sure if it's a good idea. 
Would be nice to see all the modules in app class.
I think the app module is still about common dependencies only and there is no place for other modules

app/src/main/java/solendsdkfacade/SolendSdkFacade.kt
```diff
@@ -0,0 +1,104 @@
...
+import solendsdkfacade.model.SolendMethodResultSuccess
+import solendsdkfacade.model.SolendMethodResultSuccess.SolendDepositTransactions
+import solendsdkfacade.model.SolendMethodResultSuccess.SolendWithdrawTransactions
+import timber.log.Timber
+import kotlinx.coroutines.withContext
+
+private const val TAG = "P2PSdkFacade"
+
+class SolendSdkFacade(
+    private val p2pSdk: P2pSdk,
```
1sakov, line 21: Strange naming.
We are moving from P2P, so let's make some universal name. 
SolendSdk or smth like that

app/src/main/java/solendsdkfacade/mapper/SolendMethodResultMapper.kt
```diff
@@ -0,0 +1,31 @@
...
+import solendsdkfacade.model.SolendMethodResultError
+import solendsdkfacade.model.SolendMethodResultSuccess
+import timber.log.Timber
+
+class SolendMethodResultMapper(
+    // public due to inline
+    val gson: Gson
+) {
+
+    inline fun <reified SuccessType : SolendMethodResultSuccess> fromSdk(sdkResult: String): SolendResult<SuccessType> {
```
1sakov, line 14: I think there is no profit from inlining this function.
That's ok for us that we will call a reference to this function and make the operation here.

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/AppModule.kt
gslevinkov: it's really convenient, now we need to add new module only one place and then it would be automatically added into dependency graph unit test

app/src/main/java/solendsdkfacade/SolendSdkFacade.kt
gslevinkov: I don't have any choice now, we need to use naming provided from JNI Rust interface. Renaming in process.

app/src/main/java/solendsdkfacade/mapper/SolendMethodResultMapper.kt
gslevinkov: well, it's for the reified type here to remove big `SolendMethodResultSuccess::class.java` in the parameter

davran1001 (approved):

kabramovp2p (approved):

1sakov (approved):

app/src/main/java/org/p2p/wallet/sdk/facade/model/SolendMethodResultSuccess.kt
```diff
@@ -0,0 +1,63 @@
...
+        @SerializedName("market_info")
+        val deposits: List<SolendUserDeposit>
+    ) : SolendMethodResultSuccess
+
+    class SolendUserDeposit(
+        @SerializedName("depositedAmount")
+        val depositedAmount: BigDecimal,
+        @SerializedName("symbol")
+        val depositTokenSymbol: String
+    ) : SolendMethodResultSuccess
```
1sakov, line 62: @gslevinkov can we add `Response` suffix to all these models?

app/src/main/java/org/p2p/wallet/sdk/facade/SolendSdkFacade.kt
```diff
@@ -0,0 +1,211 @@
...
+                owner_address = ownerAddress.value,
+                environment = solendEnvironment.sdkValue,
+                lendng_market_address = lendingMarketAddress.orEmpty(),
+                blockhash = currentBlockhash,
+                free_transactions_count = remainingFreeTransactionsCount,
+                need_to_use_relay = payFeeWithRelay,
+                pay_fee_in_token = feePayerToken?.let { gson.toJson(it) }.orEmpty(),
+                fee_payer_address = realFeePayerAddress.value
+            )
+            handleSdkResultWithHandler<SolendWithdrawTransactions>(result)
```
1sakov, line 104: Why do you pass the result to the handler? 
I'd prefer catching the result/error in a traditional way with try/catch in presenter. 
This will be getting more complex every time

app/src/main/java/org/p2p/wallet/sdk/facade/model/SolendMethodResultError.kt
```diff
@@ -0,0 +1,8 @@
+package org.p2p.wallet.sdk.facade.model
+
+import com.google.gson.annotations.SerializedName
+
+data class SolendMethodResultError(
```
1sakov, line 5: `SolendMethodResultException` since it's throwable

app/src/main/java/org/p2p/wallet/sdk/facade/model/SolendMethodResultSuccess.kt
```diff
@@ -0,0 +1,63 @@
+package org.p2p.wallet.sdk.facade.model
+
+import com.google.gson.annotations.SerializedName
+import org.p2p.wallet.utils.Base58String
+import java.math.BigDecimal
+import java.math.BigInteger
+
+sealed interface SolendMethodResultSuccess {
```
1sakov, line 8: Why do we need sealed interface for all models?
Why we cannot set the result model we want for each method?

app/src/main/java/org/p2p/wallet/sdk/facade/mapper/SolendMethodResultMapper.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    // public due to inline
+    fun <ResultType : SolendMethodResultSuccess> SolendResult<ResultType>.onErrorLog(
+        rawSdkResult: String
+    ): SolendResult<ResultType> {
+        if (error != null) {
+            Timber.i(rawSdkResult)
+            Timber.e(error)
+        }
+        return this
```
1sakov, line 27: This is redundant, we will show logs in presenter anyway

app/src/main/java/org/p2p/wallet/sdk/facade/SolendSdkFacade.kt
```diff
@@ -0,0 +1,211 @@
...
+    ): List<SolendUserDeposit> = withContext(dispatchers.io) {
+        Timber.tag(TAG).i("Method getSolendUserDeposits called")
+
+        methodResultMapper.fromSdk<SolendUserDepositsResponse>(
+            solendSdk.getSolendUserDeposits(
+                owner = userAddress.value,
+                pool = solendPoolAddress.poolAddress.value
+            )
+        )
+            .getOrThrow()
```
1sakov, line 202: this is redundant imho and the message is common for every call, that's not good as well

app/src/main/java/org/p2p/wallet/sdk/facade/SolendSdkFacade.kt
```diff
@@ -0,0 +1,211 @@
...
+            .accounts
+    }
+
+    suspend fun getSolendConfig(): SolendConfigResponse = withContext(dispatchers.io) {
+        Timber.tag(TAG).i("Method getSolendConfig called")
+
+        methodResultMapper.fromSdk<SolendConfigResponse>(
+            solendSdk.getSolendConfig(
+                environment = solendEnvironment.sdkValue
```
1sakov, line 128: this could be in 1 line and without names args.
And I would move the config model to the value before passing to resultMapper.
Please try to make it easy readable, it's new SDK, we don't know what it is and we need simplicity right now very much

app/src/main/java/org/p2p/wallet/sdk/facade/SolendSdkFacade.kt
```diff
@@ -0,0 +1,211 @@
...
+        Timber.tag(TAG).i("Method getSolendCollateralAccounts called")
+
+        methodResultMapper.fromSdk<SolendCollateralAccountsList>(
+            solendSdk.getSolendCollateralAccounts(
+                rpc_url = currentNetworkEnvironment.endpoint,
+                owner = ownerAddress.value
+            )
+        )
+            .getOrThrow()
+            .accounts
```
1sakov, line 120: Let's make a value of response and then return@withContext the response.accounts - that's easy to read

app/src/main/java/org/p2p/wallet/sdk/facade/SolendSdkFacade.kt
```diff
@@ -0,0 +1,211 @@
...
+        ownerAddress: Base58String,
+        lendingMarketAddress: String?,
+        currentBlockhash: String,
+        remainingFreeTransactionsCount: UInt,
+        payFeeWithRelay: Boolean,
+        feePayerToken: SolendFeePayerTokenData?,
+        realFeePayerAddress: Base58String
+    ) {
+        withContext(dispatchers.io) {
+            Timber.tag(TAG).i("Method createSolendDepositTransactions called")
```
1sakov, line 55: I think these logs are redundant. 
There are any blockers and this code will be reached anyway

app/src/main/java/org/p2p/wallet/sdk/facade/SolendSdkFacade.kt
```diff
@@ -0,0 +1,211 @@
...
+import org.p2p.wallet.sdk.facade.model.SolendMethodResultSuccess.SolendUserDepositByTokenResponse
+import org.p2p.wallet.sdk.facade.model.SolendMethodResultSuccess.SolendUserDepositsResponse
+import org.p2p.wallet.sdk.facade.model.SolendMethodResultSuccess.SolendWithdrawTransactions
+import org.p2p.wallet.utils.Base58String
+import timber.log.Timber
+import kotlinx.coroutines.withContext
+
+private const val TAG = "SolendSdkFacade"
+
+class SolendSdkFacade(
```
1sakov, line 29: Let's call it repository please, don't go out to new namings

app/src/main/java/org/p2p/wallet/sdk/facade/SolendPool.kt
```diff
@@ -0,0 +1,8 @@
+package org.p2p.wallet.sdk.facade
+
+import org.p2p.wallet.utils.Base58String
+import org.p2p.wallet.utils.toBase58Instance
+
+enum class SolendPool(val poolName: String, val poolAddress: Base58String) {
+    MAIN("main", "".toBase58Instance())
```
1sakov, line 7: Why poolAddress is empty?

app/src/main/java/org/p2p/wallet/sdk/facade/SolendMethodResultHandler.kt
```diff
@@ -0,0 +1,9 @@
+package org.p2p.wallet.sdk.facade
+
+import org.p2p.wallet.sdk.facade.model.SolendMethodResultError
+import org.p2p.wallet.sdk.facade.model.SolendMethodResultSuccess
+
+interface SolendMethodResultHandler {
+    fun handleResultSuccess(result: SolendMethodResultSuccess)
+    fun handleResultError(error: SolendMethodResultError)
```
1sakov, line 8: Don't really like the callbacks for catching the call's result.
It's absolutely different from our current way of making requests.

The developer should not see any difference between calling the simple REST request and SolendSDK call.
Let's try to create such experience

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/sdk/facade/model/SolendMethodResultSuccess.kt
gslevinkov: it will be my pleasure

app/src/main/java/org/p2p/wallet/sdk/facade/SolendSdkFacade.kt
gslevinkov: i agree, this hander should be used with the background processing in appScope that was planned earlier.
have we canceled it?

app/src/main/java/org/p2p/wallet/sdk/facade/model/SolendMethodResultSuccess.kt
gslevinkov: in the background processing entity we can have only one when and handle result depending on type
that's how it was planned

app/src/main/java/org/p2p/wallet/sdk/facade/mapper/SolendMethodResultMapper.kt
gslevinkov: we can miss Timber.e in presenter and in every SolendFacade call

app/src/main/java/org/p2p/wallet/sdk/facade/SolendSdkFacade.kt
gslevinkov: we need to trace which method of SolendSdk was called when diagnosing errors in Sentry

app/src/main/java/org/p2p/wallet/sdk/facade/SolendSdkFacade.kt
gslevinkov: can you show an example of your idea? haven't really catched that

app/src/main/java/org/p2p/wallet/sdk/facade/SolendSdkFacade.kt
gslevinkov: that's the idea!
it's not Repository, but just a Facade that should be used in Solend multiple repositories, we should split Solend different coupled methods in separate classes and not create garbage in one class

app/src/main/java/org/p2p/wallet/sdk/facade/SolendMethodResultHandler.kt
gslevinkov: will do

app/src/main/java/org/p2p/wallet/AppModule.kt
1sakov: It was the only entry place before as well, nah?

app/src/main/java/org/p2p/wallet/sdk/facade/SolendSdkFacade.kt
1sakov: No we didn't but we can call this function with result in app scope without handlers

app/src/main/java/org/p2p/wallet/sdk/facade/SolendSdkFacade.kt
1sakov:
```
val config = solendSdk.getSolendConfig(solendEnvironment.sdkValue)
val response = methodResultMapper.fromSdk(config)

return@withContext response.value
```

app/src/main/java/org/p2p/wallet/AppModule.kt
gslevinkov: no, we had an list of modules in the unit tests too
and we could forget to add module there

app/src/main/java/org/p2p/wallet/sdk/facade/SolendPool.kt
gslevinkov: unknown rn

1sakov (approved):

PR #809, opened 2022-09-29
PWN-5197 - flow fixes with custom + device share

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5197

## Description of Work

Flow fixes with custom + device share


app/src/main/java/org/p2p/wallet/auth/interactor/restore/UserRestoreInteractor.kt
```diff
@@ -93,6 +93,8 @@ class UserRestoreInteractor(
 
         restoreFlowDataLocalRepository.generateActualAccount(result.mnemonicPhrase.split(""))
         RestoreUserResult.RestoreSuccessful
+    } catch (web3AuthError: Web3AuthErrorResponse) {
```
gslevinkov, line 96: Web3AuthErrorResponse extends Throwable, why do separate catch?

gslevinkov (approved):

PR #815, opened 2022-10-01
PWN-5315 - Solend main screen added. UI only

1sakov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5315

## Description of Work

Added ui components/fragments and mock data
Next PR will be about adding real requests

## Screenshots (Optional)


kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/solend/ui/earn/SolendEarnFragment.kt
```diff
@@ -0,0 +1,52 @@
...
+    BaseMvpFragment<SolendEarnContract.View, SolendEarnContract.Presenter>(R.layout.fragment_solend_earn),
+    SolendEarnContract.View {
+
+    companion object {
+        fun create() = SolendEarnFragment()
+    }
+
+    override val presenter: SolendEarnContract.Presenter by inject()
+
+    private val earnAdapter: SolendEarnAdapter by lazy {
```
gslevinkov, line 26: can by unsafeLazy

app/src/main/java/org/p2p/wallet/solend/ui/earn/SolendEarnPresenter.kt
```diff
@@ -0,0 +1,47 @@
...
+import timber.log.Timber
+import java.math.BigDecimal
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+
+class SolendEarnPresenter(
+    private val context: Context
+) : BasePresenter<SolendEarnContract.View>(), SolendEarnContract.Presenter {
+
+    override fun load() {
```
gslevinkov, line 16: you can use onViewAttach

app/src/main/java/org/p2p/wallet/solend/ui/earn/adapter/SolendEarnAdapter.kt
```diff
@@ -0,0 +1,25 @@
+package org.p2p.wallet.solend.ui.earn.adapter
+
+import androidx.recyclerview.widget.RecyclerView
+import android.view.ViewGroup
+import org.p2p.wallet.solend.model.SolendDepositToken
+
+class SolendEarnAdapter : RecyclerView.Adapter<RecyclerView.ViewHolder>() {
```
gslevinkov, line 7: : RecyclerView.Adapter<RecyclerView.ViewHolder>() -> : RecyclerView.Adapter< SolanaEarnViewHolder >()

app/src/main/res/layout/fragment_solend_earn.xml
```diff
@@ -0,0 +1,63 @@
...
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:layout_weight="1"
+            android:paddingHorizontal="16dp"
+            android:paddingVertical="12dp"
+            android:text="@string/solend_deposit_to_earn"
+            android:textAppearance="@style/UiKit.TextAppearance.SemiBold.Text3" />
+
+        <TextView
+            android:id="@+id/textViewApy"
```
gslevinkov, line 35: what is apy?

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/solend/ui/earn/SolendEarnPresenter.kt
1sakov: I would propose not using attach for it. 
If we'll call presenter.load in fragment that would be more obvious and the developer should not enter the presenter to investigate what's being called in attach

app/src/main/res/layout/fragment_solend_earn.xml
1sakov: I may be wrong but it should be Average Percent Yield or smth like that

PR #817, opened 2022-10-02
PWN-5032 - Solend. Add feature toggle

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5032

## Description of Work

- fix using default value if value is not found in default config
- extract actions that should be done on app create into a separate entity
- load Solend conifg if toggle is true

### DIFF
--- a/app/src/main/java/org/p2p/wallet/solend/data/SolendConfigurationLocalRepository.kt
+++ b/app/src/main/java/org/p2p/wallet/solend/data/SolendConfigurationLocalRepository.kt
@@ -1,0 +1,25 @@
+package org.p2p.wallet.solend.data
+
+import org.p2p.wallet.sdk.facade.SolendSdkFacade
+import org.p2p.wallet.sdk.facade.model.SolendConfig
+import timber.log.Timber
+
+private const val TAG = "SolendConfigurationLocalRepository"
+
+class SolendConfigurationLocalRepository(
+    private val solendSdkFacade: SolendSdkFacade
+) : SolendConfigurationRepository {
+    private var solendConfig: SolendConfig? = null
+
+    override suspend fun loadSolendConfiguration() {
+        this.solendConfig = solendSdkFacade.getSolendConfig().solendConfig
+    }
+
+    override suspend fun getSolendConfiguration(): SolendConfig {
+        if (solendConfig == null) {
+            Timber.tag(TAG).i("Solend config is not loaded. Loading now...")
+            loadSolendConfiguration()
+        }
+        return solendConfig!!
+    }
+}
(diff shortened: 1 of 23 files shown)


kabramovp2p (approved):
vrode ok

app/src/main/java/org/p2p/wallet/solend/data/SolendConfigurationLocalRepository.kt
```diff
@@ -0,0 +1,25 @@
...
+class SolendConfigurationLocalRepository(
+    private val solendSdkFacade: SolendSdkFacade
+) : SolendConfigurationRepository {
+    private var solendConfig: SolendConfig? = null
+
+    override suspend fun loadSolendConfiguration() {
+        this.solendConfig = solendSdkFacade.getSolendConfig().solendConfig
+    }
+
+    override suspend fun getSolendConfiguration(): SolendConfig {
```
1sakov, line 18: `SolendConfigResponse`

And this model should be mapped to domain after it passes upper repository

app/src/main/java/org/p2p/wallet/solend/data/SolendUserDepositsRemoteRepository.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import org.p2p.wallet.infrastructure.network.provider.TokenKeyProvider
+import org.p2p.wallet.sdk.facade.SolendSdkFacade
+import org.p2p.wallet.sdk.facade.model.SolendCollateralAccountResponse
+import org.p2p.wallet.sdk.facade.model.SolendPool
+import org.p2p.wallet.sdk.facade.model.SolendUserDepositResponse
+import org.p2p.wallet.utils.Base58String
+import org.p2p.wallet.utils.toBase58Instance
+
+class SolendUserDepositsRemoteRepository(
```
1sakov, line 11: mb?
```suggestion
class SolendDepositsRemoteRepository
```

gslevinkov: looks like that we are getting all the deposits in the Solend ecosystem instead of user ones
i can rename but add `user` in the method names

gslevinkov: Kostya is working with this class, decided not to change it

1sakov: in general this class is out of scope. 
I have made a lot of changes here already

1sakov (approved):

PR #819, opened 2022-10-03
PWN-5019 - Adding requests for solend main screen

1sakov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5019

## Description of Work

Using SolendSdk requests for fetching and showing data

## Screenshots (Optional)

Screenshot_20221004_182916


app/src/main/java/org/p2p/wallet/sdk/facade/SolendSdkFacade.kt
```diff
@@ -34,7 +33,8 @@ class SolendSdkFacade(
 ) {
 
     private val currentNetworkEnvironment: NetworkEnvironment
-        get() = networkEnvironmentManager.loadCurrentEnvironment()
+        //        get() = networkEnvironmentManager.loadCurrentEnvironment()
+        get() = NetworkEnvironment.MAINNET
```
gslevinkov, line 37: MAINNET is fixed rn?

app/src/main/java/org/p2p/wallet/solend/interactor/SolendDepositsInteractor.kt
```diff
@@ -0,0 +1,24 @@
...
+) {
+
+    suspend fun getAllDeposits(tokenSymbols: List<String>): List<SolendDepositToken> {
+        val marketsInfo = repository.getSolendMarketInfo(tokenSymbols)
+        val activeDeposits = repository.getDeposits()
+
+        return marketsInfo.mapNotNull { info ->
+            val tokenData = userLocalRepository.findTokenDataBySymbol(info.tokenSymbol) ?: return@mapNotNull null
+            val activeDeposit = activeDeposits.find { it.depositTokenSymbol == info.tokenSymbol }
+            mapper.toDepositToken(tokenData, info, activeDeposit)
```
gslevinkov, line 21: interactor is not responsible for mapping network entities to domain entities
put mapping in the SolendDepositsRepository

app/src/main/java/org/p2p/wallet/solend/model/SolendDomainMapper.kt
```diff
@@ -0,0 +1,29 @@
+package org.p2p.wallet.solend.model
+
+import org.p2p.wallet.user.model.TokenData
+
+class SolendDomainMapper {
```
gslevinkov, line 5: SolendDepositTokenMapper?

app/src/main/java/org/p2p/wallet/solend/repository/SolendDepositsRemoteRepository.kt
```diff
@@ -0,0 +1,36 @@
...
+    private val ownerAddressProvider: TokenKeyProvider,
+    private val mapper: SolendDepositMapper
+) : SolendDepositsRepository {
+
+    private val currentSolendPool = SolendPool.MAIN
+
+    private val ownerAddress: Base58String
+        get() = ownerAddressProvider.publicKey.toBase58Instance()
+
+    override suspend fun getDeposits(): List<SolendUserDeposit> = try {
```
gslevinkov, line 24: getUserDeposits()

app/src/main/java/org/p2p/wallet/solend/ui/earn/SolendEarnPresenter.kt
```diff
@@ -2,46 +2,45 @@ package org.p2p.wallet.solend.ui.earn
...
                 view?.showLoading(isLoading = false)
             }
         }
     }
+
+    override fun refresh() {
+        view?.showRefreshing(isRefreshing = true)
+        launch {
+            try {
+                val deposits = solendDepositsInteractor.getAllDeposits(COLLATERAL_ACCOUNTS)
```
gslevinkov, line 36: maybe COLLATERAL_ACCOUNTS - is a domain entity? it's about domain logic, not presentation

app/src/test/java/org/p2p/wallet/sdk/facade/mapper/SolendMethodResultMapperTest.kt
```diff
@@ -0,0 +1,108 @@
...
+import com.google.gson.Gson
+import com.google.gson.JsonSyntaxException
+import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.api.Assertions.assertThatThrownBy
+import org.intellij.lang.annotations.Language
+import org.junit.Test
+import org.p2p.wallet.sdk.facade.model.SolendMarketInformationResponse
+import org.p2p.wallet.sdk.facade.model.SolendMethodResultException
+
+class SolendMethodResultMapperTest {
```
gslevinkov, line 12: wow, test master

gslevinkov (approved):

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/sdk/facade/SolendSdkFacade.kt
1sakov: this is the only one which works for now

app/src/main/java/org/p2p/wallet/solend/interactor/SolendDepositsInteractor.kt
1sakov: This is not network entities, these are merging several domain entities to single domain entity. 
And I was really struggling about the place of mapping. 

I am not sure if we need to make the repository method be complex with several remote requests. 

What do you think guys?

gslevinkov: i think if domain requests complex domain entity - it's repository responsibility to make all the needed requests and build the entity.
just like with tables - if we want multiple complex data - the repository makes all the requests in database and build it in the mapper

1sakov: okay, let me try to move it in repo

app/src/main/java/org/p2p/wallet/solend/ui/earn/SolendEarnPresenter.kt
1sakov: I think user should know entering presenter what accounts should be visible in that particalar fragment and control it from there. 
So I assume the accounts list is more as a parameter from a user but not as a result of business logic

app/src/main/java/org/p2p/wallet/solend/interactor/SolendDepositsInteractor.kt
1sakov: done

PR #834, opened 2022-10-06
PWN-5426 - fix pin confirm stackoverflow bug

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5426

## Description of Work

- make crypto more readable without changing logic
- remove recursion


solanaj/src/main/java/org/p2p/solanaj/utils/crypto/Pbkdf2HashGenerator.kt
```diff
@@ -0,0 +1,56 @@
...
+
+private const val PBKDF2_ITERATIONS = 5000
+private const val PBKDF2_DEFAULT_LENGTH_BYTES = 64
+
+class Pbkdf2Hash(
+    val hashResultInHex: String,
+    val hashSalt: ByteArray
+)
+
+class Pbkdf2HashGenerator {
```
gslevinkov, line 20: same logic as in HashingUtils

except now we can mock this hashing function and see what's its doing inside

solanaj/src/main/java/org/p2p/solanaj/utils/crypto/Pbkdf2HashGenerator.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    fun generateHashWithRandomSalt(data: String): Pbkdf2Hash {
+        require(data.isNotBlank())
+
```
gslevinkov, line 24:
```suggestion
```

PR #838, opened 2022-10-06
PWN-5317 - show error state if there is no marketsInfo but we have deposits

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5317

## Description of Work

Show error state if there is no marketsInfo but we have deposits

## Screenshot

screencast 2022-10-06 15-44-49


app/src/main/java/org/p2p/wallet/solend/repository/SolendDepositsRemoteRepository.kt
```diff
@@ -24,7 +24,12 @@ class SolendDepositsRemoteRepository(
         get() = ownerAddressProvider.publicKey.toBase58Instance()
 
     override suspend fun getUserDeposits(tokenSymbols: List<String>): List<SolendDepositToken> {
-        val marketsInfo = getSolendMarketInfo(tokenSymbols)
+        val marketsInfo = try {
+            getSolendMarketInfo(tokenSymbols)
+        } catch (e: Throwable) {
+            Timber.w(e, "Error loading user marketsInfo")
```
gslevinkov, line 30: Timber.e?

app/src/main/java/org/p2p/wallet/solend/repository/SolendDepositsRemoteRepository.kt
```diff
@@ -34,11 +39,23 @@ class SolendDepositsRemoteRepository(
             emptyList()
         }
 
-        return marketsInfo.mapNotNull { info ->
-            val tokenData = userLocalRepository.findTokenDataBySymbol(info.tokenSymbol) ?: return@mapNotNull null
-            val tokenPrice = userLocalRepository.getPriceByToken(info.tokenSymbol) ?: return@mapNotNull null
-            val activeDeposit = deposits.find { it.depositTokenSymbol == info.tokenSymbol }
-            mapper.fromNetwork(tokenData, tokenPrice, info, activeDeposit)
+        if (marketsInfo.isEmpty() && deposits.isNotEmpty()) {
```
gslevinkov, line 42: can you extract this if condition into a variable?

gslevinkov, line 42: return if

app/src/main/java/org/p2p/wallet/solend/repository/SolendDepositsRemoteRepository.kt
```diff
@@ -34,11 +39,23 @@ class SolendDepositsRemoteRepository(
...
+            return deposits.mapNotNull { deposit ->
+                val tokenData = userLocalRepository.findTokenDataBySymbol(
+                    deposit.depositTokenSymbol
+                ) ?: return@mapNotNull null
+                val tokenPrice = userLocalRepository.getPriceByToken(
+                    deposit.depositTokenSymbol
+                ) ?: return@mapNotNull null
+                mapper.fromNetwork(tokenData, tokenPrice, null, deposit)
```
gslevinkov, line 50: named arg

app/src/main/java/org/p2p/wallet/solend/repository/SolendDepositsRemoteRepository.kt
```diff
@@ -34,11 +39,23 @@ class SolendDepositsRemoteRepository(
...
-        return marketsInfo.mapNotNull { info ->
-            val tokenData = userLocalRepository.findTokenDataBySymbol(info.tokenSymbol) ?: return@mapNotNull null
-            val tokenPrice = userLocalRepository.getPriceByToken(info.tokenSymbol) ?: return@mapNotNull null
-            val activeDeposit = deposits.find { it.depositTokenSymbol == info.tokenSymbol }
-            mapper.fromNetwork(tokenData, tokenPrice, info, activeDeposit)
+        if (marketsInfo.isEmpty() && deposits.isNotEmpty()) {
+            return deposits.mapNotNull { deposit ->
+                val tokenData = userLocalRepository.findTokenDataBySymbol(
+                    deposit.depositTokenSymbol
+                ) ?: return@mapNotNull null
```
gslevinkov, line 46: please format lines the same way as you do in else branch

app/src/main/java/org/p2p/wallet/solend/ui/earn/SolendEarnPresenter.kt
```diff
@@ -94,17 +99,25 @@ class SolendEarnPresenter(
...
         when {
+            newDeposits.any { it.supplyInterest == null } -> {
+                showDepositsWidgetError()
+                view?.setRatesErrorVisibility(isVisible = true)
+            }
             newDeposits.any { it is SolendDepositToken.Active } -> {
                 val activeDeposits = newDeposits.filterIsInstance<SolendDepositToken.Active>()
                 val depositBalance = activeDeposits.sumOf { it.usdAmount }
-                val totalYearBalance = activeDeposits.sumOf { it.usdAmount / BigDecimal(100) * it.supplyInterest }
+                val totalYearBalance = activeDeposits.sumOf {
```
gslevinkov, line 109: are we sure that calculating stuff should be done in Presenter? :)

app/src/main/java/org/p2p/wallet/solend/ui/earn/SolendEarnPresenter.kt
```diff
@@ -77,10 +78,14 @@ class SolendEarnPresenter(
...
     override fun onDepositTokenClicked(deposit: SolendDepositToken) {
-        view?.showDepositTopUp(deposit)
+        if (!blockedErrorState) {
+            view?.showDepositTopUp(deposit)
+        }
     }
 
     private fun showDepositsWidgetError() {
+        timerJob?.cancel()
+        blockedErrorState = true
```
gslevinkov, line 88: blockedErrorState  -> widgetErrorState

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/solend/ui/earn/SolendEarnFragment.kt
```diff
@@ -65,6 +66,13 @@ class SolendEarnFragment :
         }
     }
 
+    override fun setRatesErrorVisibility(isVisible: Boolean) {
+        binding.root.post {
+            TransitionManager.beginDelayedTransition(binding.root)
```
1sakov, line 71: Should be set after the visibility change.
The animation will not freeze then and will be more smooth

1sakov (changes requested):
Are you sure we need to set error state for widget as well if we don't have market info?
I think it should not affect on the widget and user should be able to navigate to his deposits. 

Talk to feature owner please.

app/src/main/java/org/p2p/wallet/solend/ui/earn/SolendEarnPresenter.kt
kabramovp2p: no it is block because we block clicks =)

app/src/main/java/org/p2p/wallet/solend/ui/earn/SolendEarnPresenter.kt
kabramovp2p: where do you want to do this?

kabramovp2p: it is calculation for widget ticker

1sakov (approved):

PR #844, opened 2022-10-07
PWN-5457 - fix crash

davran1001:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5457


app/src/main/java/org/p2p/wallet/infrastructure/InfrastructureModule.kt
```diff
@@ -85,11 +85,14 @@ object InfrastructureModule : InjectionModule {
         // TODO PWN-5418 - extract data to separate prefs from org.p2p.wallet.prefs
         factory {
             val sharedPreferences: SharedPreferences = get()
+
+            val prefsName = "${androidContext().packageName}.account_prefs"
```
davran1001, line 89: we used different prefs for KeystoreWrapper, it's cause of crash

kabramovp2p (approved):

PR #848, opened 2022-10-10
PWN-3713 - latest release changes

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-3713

## Description of Work

Latest release (env / verifiers / values)  changes


app/src/main/java/org/p2p/wallet/infrastructure/network/environment/NetworkServicesUrlProvider.kt
```diff
@@ -50,11 +49,11 @@ class NetworkServicesUrlProvider(
         ).orEmpty()
         val verifier = sharedPreferences.getString(
             KEY_TORUS_BASE_VERIFIER,
-            context.getString(R.string.torusFeatureVerifier)
+            context.getString(R.string.torusReleaseVerifier)
         ).orEmpty()
         val subVerifier = sharedPreferences.getString(
             KEY_TORUS_BASE_SUB_VERIFIER,
-            emptyString()
+            context.getString(R.string.torusReleaseSubVerifier)
```
1sakov, line 56: Maybe we should pass here the certain data according to the build as well?
If DEBUG it's empty, if not, we are passing release verifier mb?

kabramovp2p: we can
but in debug we can change all this data from debug menu

PR #859, opened 2022-10-13
NO-TICKET - Сonsequences after MERGE with RESTORE

davran1001:


kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/auth/ui/phone/PhoneNumberEnterPresenter.kt
```diff
@@ -134,6 +134,9 @@ class PhoneNumberEnterPresenter(
             is GatewayHandledState.ToastError -> {
                 view?.showUiKitSnackBar(gatewayHandledResult.message)
             }
+            else -> {
+                // Do nothing
```
kabramovp2p, line 138: Unit ?

app/src/main/java/org/p2p/wallet/utils/NetworkExtensions.kt
```diff
@@ -14,6 +15,13 @@ fun Throwable?.getErrorMessage(resourcesProvider: ResourcesProvider): String =
         else -> this?.message ?: resourcesProvider.getString(R.string.error_general_message)
     }
 
+fun Throwable?.getErrorMessage(context: Context): String =
```
gslevinkov, line 18: we've deleted getErrorMessage with context so we need to use only ResourcesProvider

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/utils/NetworkExtensions.kt
davran1001: if we do this, we need to inject resourceProvider for base fragment and base activity

PR #903, opened 2022-10-26
PWN-5388 - terms update

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5388

## Description of Work

p2p_privacy_policy.pdf & p2p_terms_of_service.pdf files updated + Terms of Use -> Terms of Service


1sakov:
@kabramovp2p I can't see where you get the pdf files which you uploaded.
If I remember correctly we don't have any places right now where we show the button which opens terms of service

kabramovp2p:
> @kabramovp2p I can't see where you get the pdf files which you uploaded. If I remember correctly we don't have any places right now where we show the button which opens terms of service and looks like we need to add it

```
textViewTermsAndPolicy.apply {
                text = SpanUtils.buildTermsAndPolicyText(
                    context = requireContext(),
                    onTermsClick = { presenter.onTermsClick() },
                    onPolicyClick = { presenter.onPolicyClick() }
                )
                movementMethod = LinkMovementMethod.getInstance()
            }
```
NewOnboardingFragment

1sakov (approved):

PR #909, opened 2022-10-26
PWN-5669 - Keystore crash 1.14.1 to develop

gslevinkov:
QA requested the fix in the develop **now**


kabramovp2p (approved):
be careful with keystores and prefs :(

1sakov (approved):

PR #916, opened 2022-10-27
PWN-5712, PWN-5720 - clear onCreateActions after execute

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5712

## Description of Work

PWN-5712 - clear onCreateActions after execute them once, also fixes PWN-5720

### DIFF
--- a/app/src/main/java/org/p2p/wallet/home/MainFragment.kt
+++ b/app/src/main/java/org/p2p/wallet/home/MainFragment.kt
@@ -53,7 +53,7 @@ class MainFragment : BaseFragment(R.layout.fragment_main), MainTabsSwitcher, Cen
                 .withArgs(ARG_MAIN_FRAGMENT_ACTIONS to actions)
     }
 
-    private val onCreateActions: Array<MainFragmentOnCreateAction> by args(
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/p2p/wallet/home/MainFragment.kt
```diff
@@ -79,7 +79,10 @@ class MainFragment : BaseFragment(R.layout.fragment_main), MainTabsSwitcher, Cen
         }
         deeplinksManager.handleSavedDeeplinkIntent()
 
-        onCreateActions.forEach(::doOnCreateAction)
+        if (onCreateActions.isNotEmpty()) {
```
gslevinkov, line 82: no need for that, forEach won't invoke if array is empty

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/home/MainFragment.kt
kabramovp2p: i know but i do not want to call all this block if we are empty like onCreateActions = emptyArray()

1sakov (approved):

PR #930, opened 2022-10-30
PWN-5740 - rename Amplitude events

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5740


1sakov:
too much changes.
I assume, the target branch should be release?

gslevinkov:
updated

PR #933, opened 2022-10-30
PWN-5752 - update action panel inside the token details

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5752

## Description of Work

- refactor action buttons into custom view
- use custom view on token history screen
- hide BUY button if it's not SOL/USDC


1sakov:
Cannot get what did you do in this PR?
The task is about username in details but you have different changes here

gslevinkov:
> Cannot get what did you do in this PR? The task is about username in details but you have different changes here

sorry, wrong task linked, gotta fix that

1sakov (approved):

PR #947, opened 2022-11-09
PWN-5885 - Transitive swap with inter token btc or eth

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5885

## Description of Work

implemented with:
https://github.com/p2p-org/OrcaSwapSwift/pull/7/files
https://github.com/p2p-org/p2p-wallet-ios/pull/687/files
@1sakov @isakovch please check

### DIFF
--- a/app/src/main/java/org/p2p/wallet/swap/interactor/orca/OrcaPoolInteractor.kt
+++ b/app/src/main/java/org/p2p/wallet/swap/interactor/orca/OrcaPoolInteractor.kt
@@ -3,7 +3,6 @@ package org.p2p.wallet.swap.interactor.orca
 import org.p2p.wallet.home.model.Token
 import org.p2p.wallet.home.model.TokenComparator
 import org.p2p.wallet.infrastructure.network.provider.TokenKeyProvider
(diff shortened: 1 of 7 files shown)


1sakov (approved):
we are okay here. 
I think fees should not be changed.

app/src/main/java/org/p2p/wallet/swap/interactor/orca/OrcaPoolInteractor.kt
```diff
@@ -100,7 +100,7 @@ class OrcaPoolInteractor(
         // The transaction might be rejected because the input amount and output amount of intermediary token (wstETH) is too small
         // To temporarily fix this issue, prefers direct route or transitive route without ETH, BTC
         val directSwapPool = sortedPoolsPair.firstOrNull { it.size == 1 }
-        return directSwapPool ?: sortedPoolsPair.first()
+        return directSwapPool ?: poolsPairs.first()
```
1sakov, line 103: let's use firstOrNull, since the method returns nullable object

PR #952, opened 2022-11-09
PWN-5819 - Remove empty balance warning if username is found

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5819

## Description of Work

- no empty balance account warning if username is found


app/src/main/java/org/p2p/wallet/send/ui/search/adapter/SearchAdapter.kt
```diff
@@ -13,12 +13,29 @@ class SearchAdapter(
 
     private val data = mutableListOf<SearchResult>()
 
-    /* TODO: Will add diff later */
```
gslevinkov, line 16: later is NOW

app/src/main/java/org/p2p/wallet/send/ui/search/adapter/SearchAdapter.kt
```diff
@@ -13,12 +13,29 @@ class SearchAdapter(
...
+        private val oldList: List<SearchResult>,
+        private val newList: List<SearchResult>
+    ) : DiffUtil.Callback() {
+
+        override fun getOldListSize(): Int = oldList.size
+
+        override fun getNewListSize(): Int = newList.size
+
+        override fun areContentsTheSame(oldItemPosition: Int, newItemPosition: Int): Boolean {
+            return oldList[oldItemPosition] == newList[newItemPosition]
```
1sakov, line 26: This is not correct. 
You should add compare for balance separately

app/src/main/java/org/p2p/wallet/send/ui/main/AccountFeeView.kt
```diff
@@ -41,7 +41,7 @@ class AccountFeeView @JvmOverloads constructor(
         with(binding) {
             fillUsdFee(feeUsd)
 
-            accountImageView.setBackgroundResource(R.drawable.bg_error_rounded)
+            accountImageView.setBackgroundResource(R.drawable.bg_error_rounded_12)
```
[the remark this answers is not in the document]
gslevinkov: will do

app/src/main/java/org/p2p/wallet/send/ui/search/adapter/SearchAdapter.kt
```diff
@@ -13,12 +13,29 @@ class SearchAdapter(
...
+        }
+
+        override fun areItemsTheSame(oldItemPosition: Int, newItemPosition: Int): Boolean {
+            return oldList[oldItemPosition] === newList[newItemPosition]
```
1sakov, line 30: this should compare addresses only, since they are unique.
Otherwise this will return false and the item will be replaced by a new one

app/src/main/java/org/p2p/wallet/send/ui/search/adapter/SearchAdapter.kt
gslevinkov: why so? they are all data classes

app/src/main/java/org/p2p/wallet/send/ui/search/adapter/SearchViewHolder.kt
```diff
@@ -20,41 +23,53 @@ class SearchViewHolder(
...
+                imageResource = R.drawable.ic_wallet_gray
+            }
+
+            Glide.with(root)
+                .load(imageResource)
+                .circleCrop()
+                .into(walletImageView)
+
+            topTextView.text = item.username
+            bottomTextView.withTextOrGone(item.addressState.address.cutEnd())
```
1sakov, line 56: I think it's simpler to remove the brackets for better readability. 
the function is especially created as `infix` to use it without brackets

what do you think?
`bottomTextView withTextOrGone item.addressState.address.cutEnd()`

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/send/ui/search/adapter/SearchViewHolder.kt
gslevinkov: i even don't like this method, it created a bug for me
but i'll remove brackets only for you ;)

1sakov (approved):

davran1001 (approved):

PR #953, opened 2022-11-09
PWN-5822 PWN-5739 - Username fixes

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5822
https://p2pvalidator.atlassian.net/browse/PWN-5739

## Description of Work

- remove snackbar on pincode when creating a wallet
- remove crash when fast clicking on reserve username button


1sakov (approved):

app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinPresenter.kt
```diff
@@ -41,6 +41,13 @@ class NewCreatePinPresenter(
     private var pinMode = PinMode.CREATE
     private var navigateBackOnBackPressed = false
 
+    override fun attach(view: NewCreatePinContract.View) {
+        super.attach(view)
+        if (onboardingInteractor.currentFlow == OnboardingFlow.CreateWallet) {
+            view.showUiKitSnackBar(messageResId = R.string.auth_create_wallet_introduction)
+        }
```
1sakov, line 48: that's not looking as a best solution but it's not critical. 
the interactor should not persist the current state

gslevinkov: yeah, looks like a repository for me. i'll try and create a ticket as a tech-debt for this

PR #964, opened 2022-11-12
PWN-5938 - Cancel flow when refetching tokens on home

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5938

## Description of Work

When refetching data on HomeFragment, database is cleared and flow recieves `[]` which leads to empty state. Fixed with duck-tape, but can be refactored by removing flow and rewriting to just fetch methods without subscription.


kabramovp2p (approved):

gslevinkov:
added a new commit, so it's up to you to choose a fix for this task
- duck-taped .cancel for Flow
- refactoring to use simple methods instead of observer

app/src/main/java/org/p2p/wallet/user/interactor/UserInteractor.kt
```diff
@@ -33,6 +34,11 @@ class UserInteractor(
 
     fun getUserTokensFlow(): Flow<List<Token.Active>> =
         mainLocalRepository.getTokensFlow()
+            // remove SOL if it's zero
+            .map { userTokens ->
+                userTokens.filterNot { it.isSOL && it.isZero }
```
1sakov, line 39: we shouldn't remove sol

app/src/main/java/org/p2p/wallet/home/repository/HomeDatabaseRepository.kt
```diff
@@ -27,10 +26,7 @@ class HomeDatabaseRepository(
 
     override fun getTokensFlow(): Flow<List<Token.Active>> =
         tokenDao.getTokensFlow()
-            .map { entities ->
-                entities.map { TokenConverter.fromDatabase(it) }
-                    .sortedWith(TokenComparator())
-            }
+            .map { it.map(TokenConverter::fromDatabase) }
```
1sakov, line 29: Where did you apply sorting?

bigman212: all the sorting happens in the interactor, as it was really before

app/src/main/java/org/p2p/wallet/user/interactor/UserInteractor.kt
gslevinkov: removed

davran1001 (approved):

app/src/main/java/org/p2p/wallet/user/interactor/UserInteractor.kt
```diff
@@ -74,11 +76,12 @@ class UserInteractor(
         }
     }
 
-    suspend fun loadUserTokensAndUpdateLocal(fetchPrices: Boolean) {
+    suspend fun loadUserTokensAndUpdateLocal(fetchPrices: Boolean): List<Token.Active> {
         val newTokens = userRepository.loadUserTokens(tokenKeyProvider.publicKey, fetchPrices)
         val cachedTokens = mainLocalRepository.getUserTokens()
 
         updateLocalTokens(cachedTokens, newTokens)
+        return getUserTokens()
```
1sakov, line 84: I am not sure if it's a good idea. 
Imho it's out of this method responsibility.
I would make a separate call in presenter instead of here

gslevinkov: okay, will fix it

gslevinkov: okay, will fix it

PR #968, opened 2022-11-16
PWN-5834 - Add torus loader on restore

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5834

## Description of Work

Make loader on restore as on Create Wallet
& move AnimationView to fragment AnimationProgressFragment.kt and use it from several places to show


app/src/main/java/org/p2p/wallet/auth/ui/animationscreen/AnimationProgressFragment.kt
```diff
@@ -0,0 +1,122 @@
...
+                fragmentManager.beginTransaction()
+                    .setCustomAnimations(R.anim.fade_in, R.anim.fade_out)
+                    .add(
+                        android.R.id.content,
+                        create(loadingTitleRes),
+                        fragmentTag
+                    )
+                    .commit()
+            }
+        }
```
1sakov, line 50: I am pretty sure we can do better. 
For example launch the fragment in the root activity and hide there as well. 
We can remove static boolean fields from here as well.


Anyway this looks not good

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/auth/ui/animationscreen/AnimationProgressFragment.kt
```diff
@@ -0,0 +1,122 @@
...
+            }
+            .launchIn(viewLifecycleOwner.lifecycleScope)
+    }
+
+    override fun onDestroyView() {
+        creationProgressJob?.cancel()
+        super.onDestroyView()
+    }
+
+    data class TimerState(
```
gslevinkov, line 118: put classes to the top please

gslevinkov, line 118: put classes to the top please

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/auth/ui/animationscreen/AnimationProgressFragment.kt
kabramovp2p: sorry but not =(
because when we call this method check in fragment manager that there is already Fragment exists not working as Fragment manager creating it toooooooo long
so if we call show some times faster then it is 
we will have 2-3 fragments crated and added =(

i spend some time on fixing this 

if you or @gslevinkov have some clear and good solution i will implement this =( but FragmentManager does not have good check to find fragment same time it is added to FragmentTransaction.... =(

app/src/main/java/org/p2p/wallet/auth/ui/animationscreen/AnimationProgressFragment.kt
1sakov: I don't why we do this, but all classes should be in a separate file. 
Fragment is about fragment, not state

kabramovp2p: ok will move it out 

it was in fragment as it is not needed in other places so it is like an a private inner class

that used only in this fragment

PR #980, opened 2022-11-21
PWN-6034 - Fix bug with keyaboard

davran1001:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6034


kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/auth/ui/phone/PhoneNumberEnterFragment.kt
```diff
@@ -157,10 +157,10 @@ class PhoneNumberEnterFragment :
 
     private fun setOnResultListener() {
         requireActivity().supportFragmentManager.setFragmentResultListener(
-            REQUEST_KEY,
-            viewLifecycleOwner
+            REQUEST_KEY, viewLifecycleOwner
```
1sakov, line 160: leave it on two as a new line params

1sakov (approved):

app/src/main/java/org/p2p/wallet/auth/ui/phone/PhoneNumberEnterContract.kt
```diff
@@ -24,7 +24,7 @@ interface PhoneNumberEnterContract {
     interface Presenter : MvpPresenter<View> {
         fun onCountryCodeChanged(newCountryCode: String)
         fun onPhoneChanged(phoneNumber: String)
-        fun onCountryCodeChanged(newCountry: CountryCode)
+        fun onCountryCodeChanged(newCountry: CountryCode?)
```
gslevinkov, line 27: why nullable?

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/auth/ui/phone/PhoneNumberEnterContract.kt
davran1001: special case when we close country picker i send null to open keyboard

PR #997, opened 2022-12-02
PWN-6237 - New Send. Main screen layout and toolbar

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6237

## Description of Work

- only toolbar and fragment to work with

### DIFF
--- a/app/src/main/java/org/p2p/wallet/home/HomeModule.kt
+++ b/app/src/main/java/org/p2p/wallet/home/HomeModule.kt
@@ -16,6 +16,8 @@ import org.p2p.wallet.home.ui.new.NewSelectTokenContract
 import org.p2p.wallet.home.ui.new.NewSelectTokenPresenter
 import org.p2p.wallet.home.ui.select.SelectTokenContract
 import org.p2p.wallet.home.ui.select.SelectTokenPresenter
+import org.p2p.wallet.newsend.NewSendContract
+import org.p2p.wallet.newsend.NewSendPresenter
 import org.p2p.wallet.receive.list.TokenListContract
 import org.p2p.wallet.receive.list.TokenListPresenter
 import org.p2p.wallet.receive.network.ReceiveNetworkTypeContract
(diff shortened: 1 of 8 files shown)


kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/home/HomeModule.kt
```diff
@@ -108,6 +110,30 @@ object HomeModule : InjectionModule {
...
+        includes(createSendModule())
+    }
+
+    private fun createSendModule(): Module = module {
+        factory<SelectTokenContract.Presenter> { (tokens: List<Token>) ->
+            SelectTokenPresenter(tokens)
+        }
+        factory<SearchContract.Presenter> { (usernames: List<SearchResult>) ->
+            SearchPresenter(usernames = usernames, searchInteractor = get(), usernameDomainFeatureToggle = get())
+        }
```
1sakov, line 136: maybe it's time to create a separate module for Send feature?

app/src/main/java/org/p2p/wallet/newsend/NewSendFragment.kt
```diff
@@ -0,0 +1,41 @@
...
+import org.p2p.wallet.utils.viewbinding.viewBinding
+import org.p2p.wallet.utils.withArgs
+
+class NewSendFragment :
+    BaseMvpFragment<NewSendContract.View, NewSendContract.Presenter>(R.layout.fragment_send_new),
+    NewSendContract.View {
+
+    companion object {
+        private const val ARG_RECIPIENT_ADDRESS = "ARG_RECIPIENT_ADDRESS"
+        private const val ARG_RECIPIENT_USERNAME = "ARG_RECIPIENT_USERNAME"
```
1sakov, line 20: move it above the class

app/src/main/java/org/p2p/wallet/home/HomeModule.kt
gslevinkov: you want to extract all the send screens to a separate module SendModule
or only new ones so it would be NewSendModule?

1sakov (approved):

PR #1009, opened 2022-12-08
PWN-6052 - do not check balance for address

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6052

## Description of Work

PWN-6052 - do not check balance for address

### DIFF
--- a/app/src/main/java/org/p2p/wallet/send/interactor/SearchInteractor.kt
+++ b/app/src/main/java/org/p2p/wallet/send/interactor/SearchInteractor.kt
@@ -1,19 +1,17 @@
 package org.p2p.wallet.send.interactor
 
+import org.p2p.core.token.Token
 import org.p2p.wallet.R
 import org.p2p.wallet.auth.username.repository.UsernameRepository
 import org.p2p.wallet.common.ResourcesProvider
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/p2p/wallet/send/interactor/SearchInteractor.kt
```diff
@@ -89,7 +85,6 @@ class SearchInteractor(
                     ),
                     canReceiveAndBuy = true
                 )
-                hasEmptyBalance -> SearchResult.EmptyBalance(addressState)
```
gslevinkov, line 92: should we remove this class at all then?

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/send/interactor/SearchInteractor.kt
kabramovp2p: we can but.... do we have 100% clarity that we do not need this at all?

isakovch (approved):

1sakov (approved):

PR #1013, opened 2022-12-09
PWN-6090 - fixes for Ui and Validations + Recipient as searchResult used

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6090

## Description of Work

Fixes for Ui and Validations + Recipient as searchResult used


app/src/main/java/org/p2p/wallet/newsend/NewSendFragment.kt
```diff
@@ -45,19 +42,18 @@ class NewSendFragment :
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
         binding.toolbar.apply {
-            title = recipientUsername ?: recipientAddress
+            title = (recipient as? SearchResult.UsernameFound)?.username
+                ?: recipient.addressState.address.cutMiddle(7)
```
gslevinkov, line 46: maybe extract this logic to presenter and then call setToolbarTitle? 
also this `7` looks nasty, const maybe?

app/src/main/java/org/p2p/wallet/newsend/NewSendFragment.kt
```diff
@@ -88,12 +84,29 @@ class NewSendFragment :
         binding.widgetSendDetails.setMaxButtonVisibility(isVisible)
     }
 
+    override fun setBottomButtonText(text: String) {
+        binding.buttonBottom.text = text
+    }
+
+    override fun setBottomButtonVisibility(isVisible: Boolean) {
```
gslevinkov, line 91: i suggest using setBottomButtonIsVisible naming due to 3 states of Visibility in Android

app/src/main/java/org/p2p/wallet/newsend/NewSendPresenter.kt
```diff
@@ -110,7 +115,68 @@ class NewSendPresenter(
             is CurrencyMode.Usd -> token.totalInUsd
             is CurrencyMode.Token -> token.total.scaleLong()
         } ?: return
-        view?.setMaxButtonVisibility(isVisible = inputAmount != totalAvailable.toString())
+        // TODO PWN-6092 check on max sum - min sum for creation account - 0.00089088 SOL
+        view?.setMaxButtonVisibility(isVisible = inputAmount == emptyString())
+    }
+
+    private fun validateAmounts(token: Token.Active) {
```
gslevinkov, line 122: how about extracting this method to a separate class SendAmountValidator? Allows to make presenter more compat and responsible for viewing logic instead of domain / validation for numbers

app/src/main/res/values/send_strings.xml
```diff
@@ -50,4 +50,14 @@
     <string name="send_switch_to_token">Tap to switch to %s</string>
     <string name="send_enter_amount">Enter Amount</string>
 
+    <string name="send_slider_text_format">Send %s %s</string>
+    <string name="send_max_warning_text_format">Max: %s %s</string>
+    <string name="send_min_warning_text_format">Min: %s %s</string>
+
+    <string name="send_fees">Fees:</string>
+    <string name="send_fees_format">Fees: %s %s</string>
```
gslevinkov, line 58: do we need to set $1%s or smth like this?

ui-kit/src/main/kotlin/org/p2p/uikit/components/UiKitSendDetailsWidget.kt
```diff
@@ -78,6 +79,11 @@ class UiKitSendDetailsWidget @JvmOverloads constructor(
         binding.textViewFreeTransactions.text = text
     }
 
+    fun setFeeProgress(isVisible: Boolean) {
+        binding.progressbarFreeTransactionsProgress.isVisible = isVisible
```
gslevinkov, line 83: progressbarFreeTransactionsProgress -> progressBarFreeTransactions

ui-kit/src/main/kotlin/org/p2p/uikit/components/UiKitSendDetailsWidget.kt
```diff
@@ -78,6 +79,11 @@ class UiKitSendDetailsWidget @JvmOverloads constructor(
         binding.textViewFreeTransactions.text = text
     }
 
+    fun setFeeProgress(isVisible: Boolean) {
```
gslevinkov, line 82: setFeeProgressVisible?

gslevinkov (approved):

app/src/main/res/values/send_strings.xml
kabramovp2p: no need here but we can do it if it is strictly 
for now when you have %s %s and $1%s $2%s getString(R.string. send_fees_format, "1", "2") will work same way

ui-kit/src/main/kotlin/org/p2p/uikit/components/UiKitSendDetailsWidget.kt
kabramovp2p: setFeeProgressVisible ? or setFeeProgressIsVisible =) ?

PR #1034, opened 2022-12-20
PWN-6092 - Send screen updates, #5

1sakov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6092 

## Description of Work

- removed feature toggle
- moved root data loading to splash screen
- launching send from token details

## Screenshots (Optional)


app/src/main/java/org/p2p/wallet/splash/SplashPresenter.kt
```diff
@@ -19,7 +22,19 @@ class SplashPresenter(
...
+            loadPricesAndBids()
+        }
+    }
+
+    private fun loadPricesAndBids() {
+        launch {
+            try {
+                userInteractor.loadAllTokensData()
+            } catch (e: Throwable) {
+                Timber.e(e, "Error loading initial tokens data")
```
kabramovp2p, line 34: maybe show error screen on error? to reload data.... i mean check if we have something in data base and if nothing and we have an error do not pass user in?

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/splash/SplashPresenter.kt
1sakov: We didn't show it before.
Right now, I am not sure if it's okat to show error on Splash screen.
He will see an empty on the main. 
Anyway let's talk separately on it

PR #1035, opened 2022-12-20
PWN-6393 - check direct addresses and show error or preselect token if it was not selected

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6393

## Description of Work

PWN-6393 - check direct addresses and show error or preselect token if it was not selected


1sakov (approved):
approved, but to complex conditions, i assume we need to refactor a little bit

PR #1047, opened 2022-12-22
PWN-6312 - Add feedback button in settings

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6312

## Description of Work

- support button in settings

### DIFF
--- a/app/src/main/java/org/p2p/wallet/home/MainFragment.kt
+++ b/app/src/main/java/org/p2p/wallet/home/MainFragment.kt
@@ -14,7 +14,9 @@ import org.koin.android.ext.android.inject
 import org.p2p.uikit.components.ScreenTab
 import org.p2p.wallet.R
 import org.p2p.wallet.auth.analytics.GeneralAnalytics
+import org.p2p.wallet.common.analytics.constants.ScreenNames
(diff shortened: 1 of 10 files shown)


1sakov (approved):

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/home/MainFragment.kt
```diff
@@ -128,6 +132,17 @@ class MainFragment : BaseFragment(R.layout.fragment_main), MainTabsSwitcher, Cen
...
     override fun navigate(clickedTab: ScreenTab) {
+        if (clickedTab == ScreenTab.FEEDBACK_SCREEN) {
+            IntercomService.showMessenger()
+            analyticsInteractor.logScreenOpenEvent(ScreenNames.Main.MAIN_FEEDBACK)
+            with(binding.bottomNavigation) {
+                post { // not working reselection on last item without post
+                    setChecked(lastSelectedItemId)
+                }
+            }
+            return
```
1sakov, line 143: what if you close intercom, does the bottom navbar refresh or it's still be checked on feedback?

gslevinkov: refreshes, feedback button is not checkable

PR #1049, opened 2022-12-22
PWN-6262 - Off-ramp. Fetching items and showing in history

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6262

## Description of Work

- reusable layout
- showing different items in history


app/src/main/java/org/p2p/wallet/sell/ui/extensions/SellTransactionDetailsBindingExtensions.kt
```diff
@@ -0,0 +1,181 @@
...
+    isAddressSolWallet: Boolean,
+    walletAddress: String,
+    onAddressCopied: () -> Unit,
+) {
+    imageViewCopy.isVisible = isAddressSolWallet
+    imageViewCopy.setOnClickListener {
+        context.copyToClipBoard(walletAddress)
+        onAddressCopied()
+    }
+}
```
1sakov, line 181: Imho, the UI rendering should be in fragment. 
It's not about static extensions.
The fragment which uses these extensions are absolutely empty, as if it has empty space for exactly this code

@kabramovp2p what do you think?

app/src/main/java/org/p2p/wallet/moonpay/ui/transaction/SellTransactionDetailsBottomSheet.kt
```diff
@@ -1,22 +1,65 @@
...
+    companion object {
+        fun show(fm: FragmentManager, details: SellTransactionDetails) {
+            SellTransactionDetailsBottomSheet()
+                .withArgs(ARG_DETAILS to details)
+                .show(fm, SelectTokenBottomSheet::javaClass.name)
+        }
+    }
 
-    override val presenter: SellTransactionDetailsContract.Presenter by inject()
+    override val presenter: SellTransactionDetailsContract.Presenter = SellTransactionDetailsPresenter()
```
1sakov, line 38: nice

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryContract.kt
```diff
@@ -3,15 +3,16 @@ package org.p2p.wallet.history.ui.history
...
 import org.p2p.wallet.history.model.HistoryTransaction
 
 interface HistoryContract {
     interface View : MvpView {
         fun showPagingState(state: PagingState)
-        fun showHistory(items: List<HistoryTransaction>)
         fun openTransactionDetailsScreen(transaction: HistoryTransaction)
         fun showRefreshing(isRefreshing: Boolean)
         fun scrollToTop()
+        fun showHistory(items: List<HistoryTransaction>, moonpayItems: List<HistoryItem.MoonpayTransactionItem>)
```
1sakov, line 15: Why you didn't unify MoonpayTransaction to `HistoryTransaction`?

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/sell/ui/extensions/SellTransactionDetailsBindingExtensions.kt
kabramovp2p: I agree to make UI related logic in fragment

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryPresenter.kt
```diff
@@ -81,15 +86,19 @@ class HistoryPresenter(
...
     private suspend fun fetchHistory(isRefresh: Boolean = false) {
         try {
+            val moonpayItems = sellInteractor.loadUserSellTransactionsDetails()
+                .map { HistoryItem.MoonpayTransactionItem(it) }
+            moonpayTransactions.clear()
+            moonpayTransactions.addAll(moonpayItems)
             val fetchedItems = historyInteractor.loadTransactions(isRefresh)
             transactions.addAll(fetchedItems)
-            view?.showHistory(transactions)
+            view?.showHistory(transactions, moonpayItems)
```
1sakov, line 95: I would split the requests into separate suspend functions and call them here. 
This function is getting complex

kabramovp2p (approved):
do not like 2 transactions lists in adapter =(

app/src/main/java/org/p2p/wallet/moonpay/ui/transaction/SellTransactionDetailsBottomSheet.kt
gslevinkov: thanks, going to fix that!

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryContract.kt
gslevinkov: HistoryTransaction has blockchain data inside, we don't have that in moonpay

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryPresenter.kt
gslevinkov: i'll do it in a separate task, history fragment is a mess that i will work on later

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryContract.kt
1sakov: I think we can remove default parameters from the sealed class and blockchain data will not be required then

app/src/main/java/org/p2p/wallet/sell/ui/extensions/SellTransactionDetailsBindingExtensions.kt
gslevinkov: how can i use the same render logic in another fragment?

PR #1052, opened 2022-12-23
PWN-6262 - Off-ramp. Bottom sheet layout 1/2

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6262


app/src/main/java/org/p2p/wallet/sell/ui/lock/SellLockedFragment.kt
```diff
@@ -26,44 +33,100 @@ class SellLockedFragment :
...
     companion object {
-        fun create(arguments: SellLockedArguments): SellLockedFragment =
-            SellLockedFragment()
-                .withArgs(ARG_SELL_LOCKED to arguments)
+        fun create(details: SellTransactionDetails): SellLockedFragment {
+            require(details.status == SellTransactionStatus.WAITING_FOR_DEPOSIT) {
+                "This fragment is used only if status == waiting for deposit"
+            }
+            return SellLockedFragment()
+                .withArgs(ARG_SELL_LOCKED to details)
```
1sakov, line 41: Weird.
Maybe you just create a function only for the determined status then?

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/sell/ui/lock/SellLockedFragment.kt
gslevinkov: it's not a sealed class, i can't pass only determined status there 
i will cast it to sealed class later, in another task, there will be a new field there

app/src/main/java/org/p2p/wallet/home/ui/main/bottomsheet/HomeActionsBottomSheet.kt
```diff
@@ -55,12 +53,6 @@ class HomeActionsBottomSheet :
         expandToFitAllContent()
```
kabramovp2p, line 53: remove it from here as lon as it is called in BaseMvpBottomSheet

app/src/main/java/org/p2p/wallet/moonpay/ui/transaction/SellTransactionDetailsBottomSheet.kt
```diff
@@ -1,22 +1,197 @@
...
+        when (action) {
+            SellTransactionDetailsButtonAction.SEND -> {
+                val recipient = SearchResult.AddressOnly(AddressState(details.receiverAddress, NetworkType.SOLANA))
+                replaceFragment(NewSendFragment.create(recipient = recipient))
+            }
+            SellTransactionDetailsButtonAction.CLOSE -> {
+                dismissAllowingStateLoss()
+            }
+            SellTransactionDetailsButtonAction.TRY_AGAIN -> {
+                // todo should be a task somewhere to implement try again feature
```
kabramovp2p, line 72: ticket number if it is exists

app/src/main/java/org/p2p/wallet/moonpay/ui/transaction/SellTransactionDetailsBottomSheet.kt
```diff
@@ -1,22 +1,197 @@
...
+            }
+        }
+    }
+
+    private fun setupViews() = with(binding.layoutDetails) {
```
kabramovp2p, line 77: big big method =(

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/home/ui/main/bottomsheet/HomeActionsBottomSheet.kt
gslevinkov: it's not called in BaseMvpBottomSheet

app/src/main/java/org/p2p/wallet/moonpay/ui/transaction/SellTransactionDetailsBottomSheet.kt
gslevinkov: i know, right? 
i can extract it to class if it helps..

1sakov (approved):

app/src/main/java/org/p2p/wallet/sell/ui/lock/SellLockedFragment.kt
1sakov: Yeah, would be nice. 
We should avoid throwing exception when creating a fragment

PR #1057, opened 2022-12-23
PWN-6538 - transaction details bottomsheet ui and layout fixes

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6538

## Description of Work

PWN-6538 - transaction details bottomsheet ui and layout fixes

### DIFF
--- a/app/src/main/java/org/p2p/wallet/common/mvp/BaseMvpBottomSheet.kt
+++ b/app/src/main/java/org/p2p/wallet/common/mvp/BaseMvpBottomSheet.kt
@@ -1,12 +1,13 @@
 package org.p2p.wallet.common.mvp
 
-import androidx.annotation.CallSuper
-import androidx.annotation.LayoutRes
-import androidx.annotation.StringRes
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import androidx.annotation.CallSuper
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/p2p/wallet/common/mvp/BaseMvpBottomSheet.kt
```diff
@@ -42,6 +44,12 @@ abstract class BaseMvpBottomSheet<V : MvpView, P : MvpPresenter<V>>(
         presenter.detach()
     }
 
+    protected fun expandToFitAllContent() {
```
gslevinkov, line 47: private

gslevinkov (approved):

1sakov (approved):

app/src/main/java/org/p2p/wallet/common/mvp/BaseMvpBottomSheet.kt
kabramovp2p: we could override it with `= Unit` to prevent expanding in future

gslevinkov: sounds like a boolean flag for me, i wouldn't ever think that i need to override method to Unit if i want to collapse bottom sheet..

PR #1058, opened 2022-12-23
PWN-6554 - provide showToast everywhere!

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6554

## Description of Work

Provide showToast in MvpView.kt and use it to show max on new Send

### DIFF
--- a/app/src/main/java/org/p2p/wallet/common/mvp/BaseFragmentContract.kt
+++ b/app/src/main/java/org/p2p/wallet/common/mvp/BaseFragmentContract.kt
@@ -3,14 +3,12 @@ package org.p2p.wallet.common.mvp
 import androidx.annotation.AnimRes
 import androidx.annotation.StringRes
 import com.google.android.material.snackbar.Snackbar
+import org.p2p.core.common.TextContainer
 
 interface BaseFragmentContract {
     fun overrideEnterAnimation(@AnimRes animation: Int)
     fun overrideExitAnimation(@AnimRes animation: Int)
-    fun showToast(
-        message: String? = null,
-        @StringRes messageResId: Int? = null
-    )
+    fun showToast(message: TextContainer)
     fun showUiKitSnackBar(
         message: String? = null,
         @StringRes messageResId: Int? = null,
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/p2p/wallet/common/mvp/BaseFragmentContract.kt
```diff
@@ -7,6 +7,10 @@ import com.google.android.material.snackbar.Snackbar
 interface BaseFragmentContract {
     fun overrideEnterAnimation(@AnimRes animation: Int)
     fun overrideExitAnimation(@AnimRes animation: Int)
+    fun showToast(
+        message: String? = null,
+        @StringRes messageResId: Int? = null
+    )
```
1sakov, line 13: use `TextContainer` instead of two nullable types

1sakov (changes requested):

gslevinkov:
can i see the design of such toast?

app/src/main/java/org/p2p/wallet/common/mvp/BaseFragmentContract.kt
kabramovp2p: Waaaat ?

kabramovp2p:
> can i see the design of such toast?

take this
image

app/src/main/java/org/p2p/wallet/common/mvp/BaseFragmentContract.kt
1sakov: You'll be able to pass 
`TextContainer(R.string.some_string)` or 
`TextContainer("Some string")`

And get the value by `getString(context)`

kabramovp2p: ohh i see
Djezd'

kabramovp2p: @1sakov fixed

1sakov (approved):

PR #1061, opened 2022-12-26
NO-TICKET - random corrections

1sakov:
NO-TICKET - Optimized sorting
NO-TICKET - Fixed hide buttons on the main screen and behavior
NO-TICKET - Removed old colors from theme

### DIFF
--- a/app/src/main/java/org/p2p/wallet/swap/interactor/orca/OrcaPoolInteractor.kt
+++ b/app/src/main/java/org/p2p/wallet/swap/interactor/orca/OrcaPoolInteractor.kt
@@ -170,14 +170,15 @@ class OrcaPoolInteractor(
     private suspend fun mapTokensForDestination(orcaTokens: List<OrcaToken>): List<Token> {
         val userTokens = userInteractor.getUserTokens()
         val publicKey = tokenKeyProvider.publicKey
-        val allTokens = orcaTokens.mapNotNull { orcaToken ->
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/p2p/wallet/swap/interactor/orca/OrcaPoolInteractor.kt
```diff
@@ -170,14 +170,15 @@ class OrcaPoolInteractor(
...
         val userTokens = userInteractor.getUserTokens()
         val publicKey = tokenKeyProvider.publicKey
-        val allTokens = orcaTokens.mapNotNull { orcaToken ->
-            val userToken = userTokens.find { it.mintAddress == orcaToken.mint }
-            if (userToken != null) {
-                userToken.takeUnless { userToken.isSOL && userToken.publicKey != publicKey }
-            } else {
-                userInteractor.findTokenData(orcaToken.mint)
+        val allTokens = orcaTokens
+            .mapNotNull { orcaToken ->
```
gslevinkov, line 174: the first call should on one line with the orcaTokens

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/swap/interactor/orca/OrcaPoolInteractor.kt
1sakov: yeah, I know the rules. 
But for this case it was hard to read. 
imho in this case this one is more readable

PR #1070, opened 2022-12-28
PWN-6277 PWN-6512 - Off ramp. Corner cases

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6277

## Description of Work

PWN-6277 - Off-ramp. remove action buttons if no internet
PWN-6277 PWN-6512 - Off-ramp. corner cases

### DIFF
--- a/app/src/main/java/org/p2p/wallet/home/ui/main/HomePresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/home/ui/main/HomePresenter.kt
@@ -96,6 +96,7 @@ class HomePresenter(
         if (state.tokens.isEmpty()) {
             initialLoadTokens()
         } else {
+            initializeActionButtons()
             handleUserTokensLoaded(state.tokens)
         }
         startPollingForTokens()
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/p2p/wallet/home/ui/main/HomePresenter.kt
```diff
@@ -96,6 +96,7 @@ class HomePresenter(
         if (state.tokens.isEmpty()) {
             initialLoadTokens()
         } else {
+            initializeActionButtons()
             handleUserTokensLoaded(state.tokens)
```
1sakov, line 100: If user is new and he has internet, he'll not see the action buttons?

gslevinkov: he will see them is everything is ok in initialLoadTokens

1sakov: I just checked that `initializeActionButtons` is being called on here.
If `initialLoadTokens` succeeded it doesn't set the buttons to adapter. 
Are you sure it will work?

1sakov (approved):

PR #1075, opened 2022-12-30
PWN-6463 - Off ramp. Handle sell transaction cancel failure

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6464

[PWN-6464]: https://p2pvalidator.atlassian.net/browse/PWN-6464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/org/p2p/wallet/sell/ui/lock/SellLockedFragment.kt
```diff
@@ -128,4 +120,23 @@ class SellLockedFragment :
...
+    override fun navigateBack() {
+        popBackStackTo(SellPayloadFragment::class)
+    }
+
+    override fun navigateBackToMain() {
+        popBackStackTo(MainFragment::class)
+    }
+
+    private fun showWarningDialog() {
+        MaterialAlertDialogBuilder(requireContext())
```
kabramovp2p, line 133: material dialog and Alert dialog has different styles, did you check that it is what we want ?

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/sell/ui/lock/SellLockedFragment.kt
gslevinkov: i wrote styles special for MaterialAlertDialog, other dialogs are not modified

kabramovp2p: thanks for that sir )

1sakov (approved):

PR #1076, opened 2023-01-01
NO-TICKET - Fix onBackPressed behavior in whole app

gslevinkov:


gslevinkov:
you can check it by checkout develop and trying to press system back on sign in screen

app/src/main/java/org/p2p/wallet/root/RootActivity.kt
```diff
@@ -104,10 +103,6 @@ class RootActivity :
         toast(message)
     }
 
-    override fun onBackPressed() {
-        popBackStack()
```
kabramovp2p, line 108: sure that we need to remove this ?

gslevinkov: again, try to debug all onBackDispatchers outside this branch and check that onBackDispatchers are not working

kabramovp2p:
> you can check it by checkout develop and trying to press system back on sign in screen

did you checked all other places that they are not broken ?

app/src/main/java/org/p2p/wallet/root/RootActivity.kt
1sakov: This will break the navigation. 
This is the behavior for the fragments that don't have dispatcher callbacks.

gslevinkov:
@1sakov will fix in another PR

PR #1103, opened 2023-01-12
PWN-6774 - transaction history filtering

NikolaiEmelianov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6774

## Description of Work

transaction history filtering


app/src/main/java/org/p2p/wallet/history/ui/history/HistoryMoonpayTransactionsMapper.kt
```diff
@@ -0,0 +1,25 @@
+package org.p2p.wallet.history.ui.history
+
+import org.p2p.wallet.common.date.toZonedDateTime
+import org.p2p.wallet.moonpay.model.SellTransaction
+import org.p2p.wallet.moonpay.serversideapi.response.SellTransactionStatus
+import org.threeten.bp.Duration
+import org.threeten.bp.ZonedDateTime
+
+class HistoryMoonpayTransactionsMapper {
```
gslevinkov, line 9: HistorySellTransactionMapper

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryMoonpayTransactionsMapper.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    fun map(transaction: List<SellTransaction>): List<SellTransaction> {
+        val now = ZonedDateTime.now()
+        return transaction
+            .filter {
+                it.status == SellTransactionStatus.WAITING_FOR_DEPOSIT || it.isNotExpired(now)
+            }
+            .sortedBy { it.metadata.createdAt.toZonedDateTime() }
```
gslevinkov, line 17:
```suggestion
        return transaction
            .filter { it.status == SellTransactionStatus.WAITING_FOR_DEPOSIT || it.isNotExpired(now) }
            .sortedBy { it.metadata.createdAt.toZonedDateTime() }
```

kabramovp2p (approved):

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/common/date/ZonedDateExtensions.kt
```diff
@@ -51,5 +51,10 @@ fun Long.toZonedDateTime(): ZonedDateTime = ZonedDateTime.ofInstant(
     ZoneId.systemDefault()
 )
 
+fun String.toZonedDateTime(): ZonedDateTime = ZonedDateTime.ofInstant(
+    Instant.parse(this),
+    ZoneId.systemDefault()
```
1sakov, line 56: this may give you an exception and the app may crash.

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryPresenter.kt
```diff
@@ -87,10 +88,9 @@ class HistoryPresenter(
     // refactor and make good code in PWN-6386
     private suspend fun fetchHistory(isRefresh: Boolean = false) {
         try {
-            if (sellInteractor.isSellAvailable()) {
-                fetchMoonpayTransactions()
-            }
+            val fetchMoonpayJob = if (sellInteractor.isSellAvailable()) launch { fetchMoonpayTransactions() } else null
             val fetchedItems = historyInteractor.loadTransactions(isRefresh)
+            fetchMoonpayJob?.join()
```
1sakov, line 93: nice!

1sakov (approved):

app/src/main/java/org/p2p/wallet/common/date/ZonedDateExtensions.kt
NikolaiEmelianov: yes it throws DateTimeParseException – if the text cannot be parsed or 
DateTimeException – if the result exceeds the supported range. Maybe add try, if catch return ""?

app/src/main/java/org/p2p/wallet/common/date/ZonedDateExtensions.kt
```diff
@@ -51,5 +52,14 @@ fun Long.toZonedDateTime(): ZonedDateTime = ZonedDateTime.ofInstant(
     ZoneId.systemDefault()
 )
 
+fun String.toZonedDateTime(): ZonedDateTime {
```
Davran331122, line 55: I am not sure, that this function should work this way. In case if string is not instance of ISO we should return null.
You can add extra function like  - `asZonedDateTimeOrNow()`

app/src/main/java/org/p2p/wallet/history/ui/history/HistorySellTransactionMapper.kt
```diff
@@ -0,0 +1,23 @@
...
+        val now = ZonedDateTime.now()
+        return transaction
+            .filter { it.status == SellTransactionStatus.WAITING_FOR_DEPOSIT || it.isNotExpired(now) }
+            .sortedBy { it.metadata.createdAt.toZonedDateTime() }
+    }
+
+    private fun SellTransaction.isNotExpired(now: ZonedDateTime): Boolean {
+        val updatedAt = metadata.updatedAt.toZonedDateTime()
+        val hoursDuration = Duration.between(updatedAt, now).toHours()
+        return hoursDuration <= 24L
```
Davran331122, line 21: move `24L` value to variable please )

Davran331122 (approved):

PR #1117, opened 2023-01-17
PWN-6783 - fix error state when it is error on counting fees

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6783

## Description of Work

PWN-6783 - fix error state when it is error on counting fees

### DIFF
--- a/app/src/main/java/org/p2p/wallet/newsend/model/NewSendButtonState.kt
+++ b/app/src/main/java/org/p2p/wallet/newsend/model/NewSendButtonState.kt
@@ -54,10 +54,6 @@ class NewSendButtonState(
             val isMinRequiredBalanceLeft = isMinRequiredBalanceLeft()
 
             return when {
-                !isAmountNotZero -> {
-                    val textContainer = TextContainer.Res(R.string.main_enter_the_amount)
-                    State.Disabled(textContainer, R.color.text_night)
-                }
                 !isFeeCalculationValid -> {
                     val textContainer = TextContainer.Res(R.string.send_cant_calculate_fees_error)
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/p2p/wallet/newsend/model/NewSendButtonState.kt
```diff
@@ -97,6 +93,10 @@ class NewSendButtonState(
                         totalAmountTextColor = R.color.text_rose
                     )
                 }
+                !isAmountNotZero -> {
```
1sakov, line 96: make it as a 2-nd condition please, otherwise it's not working

ui-kit/src/main/kotlin/org/p2p/uikit/components/UiKitSendDetailsWidget.kt
```diff
@@ -99,7 +99,11 @@ class UiKitSendDetailsWidget @JvmOverloads constructor(
 
     fun showFeeLoading(isLoading: Boolean) {
         binding.progressBarFees.isVisible = isLoading
-        binding.imageViewFeesInfo.isInvisible = isLoading
+        if (binding.textViewFee.text.isNullOrBlank()) {
+            binding.imageViewFeesInfo.isVisible = false
+        } else {
+            binding.imageViewFeesInfo.isInvisible = isLoading
+        }
```
1sakov, line 106: I am not getting this. 
looks like a hack. 
What is the problem here?

1sakov (approved):

NikolaiEmelianov (approved):

PR #1126, opened 2023-01-18
PWN-6878 - update cashout transaction after moonpay

NikolaiEmelianov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6878

## Description of Work

update cashout transaction after moonpay


app/src/main/java/org/p2p/wallet/sell/ui/payload/SellPayloadFragment.kt
```diff
@@ -56,6 +56,16 @@ class SellPayloadFragment :
         }
     }
 
+    override fun onStart() {
```
kabramovp2p, line 59: heh =(

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/sell/ui/payload/SellPayloadFragment.kt
NikolaiEmelianov: ?

gslevinkov: can we call presenter method to update data only on onResume?

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/sell/ui/payload/SellPayloadContract.kt
```diff
@@ -42,5 +42,7 @@ interface SellPayloadContract {
         fun cashOut()
         fun onTokenAmountChanged(newValue: String)
         fun onUserMaxClicked()
+        fun onStart()
+        fun onStop()
```
1sakov, line 46: Can you name this methods with more context? 
For example, `checkSellLock` or smth

app/src/main/java/org/p2p/wallet/sell/ui/payload/SellPayloadPresenter.kt
```diff
@@ -78,6 +79,7 @@ class SellPayloadPresenter(
     }
 
     private suspend fun checkForSellLock() {
+        needCheckForSellLock = false
```
1sakov, line 82: What does it mean? 
Sell Lock? Does it mean the user is blocked for selling?
Maybe you'll call it `isSellAvailable` then?

app/src/main/java/org/p2p/wallet/sell/ui/payload/SellPayloadPresenter.kt
```diff
@@ -95,6 +97,14 @@ class SellPayloadPresenter(
         }
     }
 
+    override fun onStart() {
+        launch { if (needCheckForSellLock) checkForSellLock() }
```
1sakov, line 101: You could move `if (needCheckForSellLock)` inside `checkForSellLock` function

app/src/main/java/org/p2p/wallet/sell/ui/payload/SellPayloadPresenter.kt
```diff
@@ -95,6 +97,14 @@ class SellPayloadPresenter(
...
+    }
+
+    override fun onStop() {
+        needCheckForSellLock = true
```
1sakov, line 105: I cannot get, why are you changing this value on stop. 
What is the idea?

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/sell/ui/payload/SellPayloadPresenter.kt
gslevinkov: we have a fragment `SellLocked` when the user is locked from the sell 
so we check that is Sell is locked for user, naming is ok here, i approve

app/src/main/java/org/p2p/wallet/sell/ui/payload/SellPayloadContract.kt
Davran331122: we can implement DefaultLifecycleObserver presenter contract, to have this methods as Default

Davran331122 (approved):

app/src/main/java/org/p2p/wallet/sell/ui/payload/SellPayloadContract.kt
1sakov: we don't need it, it's rare case

app/src/main/java/org/p2p/wallet/sell/ui/payload/SellPayloadPresenter.kt
NikolaiEmelianov: `checkForSellLock()`call on `attach` `org/p2p/wallet/sell/ui/payload/SellPayloadPresenter.kt:65` and have comment `// call order is important!`
also we need to call `checkForSellLock()` only after return from moonpay webView for `navigateToSellLock` `org/p2p/wallet/sell/ui/payload/SellPayloadPresenter.kt:89`

so if fragment livecycle only onStop()-> onStart() need to call `checkForSellLock`
if fragment recreated i don't want to call `checkForSellLock` twice `needCheckForSellLock` is control it

Davran331122 (approved):

1sakov (approved):

PR #1127, opened 2023-01-18
PWN-6866 - make less then string if amount is in 0...0.01 range

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6866

## Description of Work

Make less then string if amount is in 0...0.01 range

### DIFF
--- a/app/src/main/java/org/p2p/wallet/newsend/model/CalculationMode.kt
+++ b/app/src/main/java/org/p2p/wallet/newsend/model/CalculationMode.kt
@@ -6,6 +6,7 @@ import org.p2p.core.utils.emptyString
 import org.p2p.core.utils.formatToken
 import org.p2p.core.utils.formatUsd
 import org.p2p.core.utils.fromLamports
+import org.p2p.core.utils.lessThenMinValue
 import org.p2p.core.utils.orZero
 import org.p2p.core.utils.scaleLong
 import org.p2p.core.utils.toBigDecimalOrZero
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/p2p/wallet/newsend/model/CalculationMode.kt
```diff
@@ -149,15 +153,36 @@ class CalculationMode(private val sendModeProvider: SendModeProvider) {
...
+            } else {
+                tokenAmount.formatToken(token.decimals)
+            },
+            token.tokenSymbol
+        )
+    }
+
+    private fun handleCalculateUsdAmountUpdate() {
+        handleCalculationUpdate(
+            if (usdAmount.lessThenMinValue()) {
```
gslevinkov, line 179: named args please

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/newsend/model/CalculationMode.kt
```diff
@@ -149,15 +153,36 @@ class CalculationMode(private val sendModeProvider: SendModeProvider) {
...
+    }
+
+    private fun handleCalculateTokenAmountUpdate() {
+        handleCalculationUpdate(
+            if (tokenAmount.lessThenMinValue()) {
+                lessThenMinString
+            } else {
+                tokenAmount.formatToken(token.decimals)
+            },
+            token.tokenSymbol
```
1sakov, line 173: You need to make Less than zero only for USD. 
The tokens has decimals up to 9 fraction

app/src/main/java/org/p2p/wallet/newsend/model/CalculationMode.kt
1sakov:
```
val formattedUsdAmount = if (usdAmount.lessThenMinValue()) lessThenMinString else usdAmount.formatUsd()
handleCalculationUpdate(formattedUsdAmount, USD_READABLE_SYMBOL)
```

1sakov (changes requested):

Davran331122 (approved):

1sakov (approved):

PR #1154, opened 2023-01-24
PWN-6945 - toolbar title, insets extensions

NikolaiEmelianov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6945

## Description of Work

toolbar title, insets extensions

### DIFF
--- a/app/src/main/java/org/p2p/wallet/sell/ui/payload/SellPayloadFragment.kt
+++ b/app/src/main/java/org/p2p/wallet/sell/ui/payload/SellPayloadFragment.kt
@@ -5,6 +5,11 @@ import android.os.Bundle
 import android.view.View
 import org.koin.android.ext.android.inject
 import org.p2p.core.utils.hideKeyboard
+import org.p2p.core.utils.insets.appleBottomInsets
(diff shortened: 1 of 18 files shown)


app/src/main/java/org/p2p/wallet/sell/ui/payload/SellPayloadFragment.kt
```diff
@@ -111,6 +125,15 @@ class SellPayloadFragment :
         setButtonState(newState.cashOutButtonState)
     }
 
+    override fun updateToolbarTitle(tokenSymbol: String) {
```
gslevinkov, line 128: tokenSymbol is never empty, and it's always SOL

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/sell/ui/payload/SellPayloadFragment.kt
NikolaiEmelianov: always SOL now, not always SOL in the future

Davran331122 (approved):

PR #1166, opened 2023-01-25
PWN-6863 - fix visibility and set default labels

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6863

## Description of Work

PWN-6863 - fix init on start of functionality and fix visibility condition


gslevinkov (approved):

ui-kit/src/main/res/layout/widget_send_details_input.xml
```diff
@@ -41,6 +41,7 @@
         app:layout_constraintHorizontal_bias="1"
         app:layout_constraintStart_toEndOf="@id/textViewSendLabel"
         app:layout_constraintTop_toTopOf="parent"
+        app:layout_goneMarginEnd="32dp"
```
Davran331122, line 44: marginEnd or goneMarginEnd ?

Davran331122 (approved):

app/src/main/java/org/p2p/wallet/newsend/SendFeeRelayerManager.kt
```diff
@@ -44,6 +44,7 @@ class SendFeeRelayerManager(
     private lateinit var feeLimitInfo: FreeTransactionFeeLimit
     private lateinit var recipientAddress: SearchResult
     private lateinit var solToken: Token.Active
+    private var initializeCompleted = false
```
1sakov, line 47: I think we should fix the case in another way. 
This is a hack to fix but we need to find an issue

ui-kit/src/main/kotlin/org/p2p/uikit/components/UiKitSendDetailsWidget.kt
```diff
@@ -98,12 +98,9 @@ class UiKitSendDetailsWidget @JvmOverloads constructor(
...
     fun showFeeLoading(isLoading: Boolean) {
-        binding.progressBarFees.isVisible = isLoading
-        if (binding.textViewFee.text.isNullOrBlank()) {
-            binding.imageViewFeesInfo.isVisible = false
-        } else {
-            binding.imageViewFeesInfo.isInvisible = isLoading
-        }
+        val canBeVisible = binding.textViewFee.isVisible
+        binding.progressBarFees.isVisible = canBeVisible && isLoading
+        binding.imageViewFeesInfo.isVisible = canBeVisible && !isLoading
```
1sakov, line 103: this is very hacky

ui-kit/src/main/kotlin/org/p2p/uikit/components/UiKitSendDetailsWidget.kt
```diff
@@ -114,8 +111,8 @@ class UiKitSendDetailsWidget @JvmOverloads constructor(
         }
 
         handler.postDelayed(PROGRESS_DELAY_IN_MS) {
-            binding.progressBarFees.isVisible = true
-            binding.imageViewFeesInfo.isInvisible = true
+            binding.progressBarFees.isVisible = binding.textViewFee.isVisible
```
1sakov, line 114: same here

1sakov (changes requested):

ui-kit/src/main/res/layout/widget_send_details_input.xml
kabramovp2p: goneMarginEnd it is for cases when we make progressBar  GONE but we do not want to shift this view

ui-kit/src/main/kotlin/org/p2p/uikit/components/UiKitSendDetailsWidget.kt
kabramovp2p: better then before
it is just an a condition not to show view when another view is GONE
because we have progress and info view
and text
progress show shown when we load fee
info when not loading
text is a main view if we do not have text of fees we can't show progress or info

another fix can be done in layout
wrap this views in Linear layout and hide it all

1sakov: The linear layout sounds good to me

Davran331122 (approved):

1sakov (approved):

PR #1169, opened 2023-01-26
PWN-6463 - Off-ramp. Fix cancel_transaction success handle

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6463

### DIFF
--- a/app/src/main/java/org/p2p/wallet/moonpay/serversideapi/MoonpayServerSideApi.kt
+++ b/app/src/main/java/org/p2p/wallet/moonpay/serversideapi/MoonpayServerSideApi.kt
@@ -1,6 +1,7 @@
 package org.p2p.wallet.moonpay.serversideapi
 
 import org.p2p.wallet.moonpay.serversideapi.response.MoonpaySellTransactionShortResponse
+import retrofit2.Response
 import retrofit2.http.DELETE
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/p2p/wallet/moonpay/serversideapi/MoonpayServerSideApi.kt
```diff
@@ -20,5 +21,5 @@ interface MoonpayServerSideApi {
     @DELETE("api/v3/sell_transactions/{transactionId}")
     suspend fun cancelSellTransaction(
         @Path("transactionId") sellTransactionId: String
-    )
+    ): Response<Unit>
```
kabramovp2p, line 24: Response<Unit> ? strange move

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/moonpay/serversideapi/MoonpayServerSideApi.kt
gslevinkov: i know right?
retrofit stuff
204 returns empty body and he fails to parse it without this

NikolaiEmelianov (approved):

PR #1174, opened 2023-01-26
PWN-6991 - Fix history state when switching between fragments

Davran331122:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6991

Fix history state when switching between fragments


app/src/main/java/org/p2p/wallet/history/ui/token/TokenHistoryFragment.kt
```diff
@@ -186,6 +179,7 @@ class TokenHistoryFragment :
     }
 
     override fun showPagingState(newState: PagingState) {
+        Timber.tag("_____STATE").d(newState.toString())
```
Davran331122, line 182: I will remove this )

gslevinkov (approved):

kabramovp2p (approved):

PR #1182, opened 2023-01-30
PWN-7013 - delay after first request

NikolaiEmelianov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7013

## Description of Work

delay after first request

### DIFF
--- a/app/src/main/java/org/p2p/wallet/sell/ui/payload/SellPayloadPresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/sell/ui/payload/SellPayloadPresenter.kt
@@ -190,7 +190,6 @@ class SellPayloadPresenter(
         sellQuoteJob = launch {
             while (isActive) {
                 try {
-                    delay(SELL_QUOTE_REQUEST_DEBOUNCE_TIME)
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/p2p/wallet/sell/ui/payload/SellPayloadPresenter.kt
```diff
@@ -188,16 +189,18 @@ class SellPayloadPresenter(
...
-                    delay(SELL_QUOTE_REQUEST_DEBOUNCE_TIME)
                     val sellQuote = sellInteractor.getSellQuoteForSol(
                         solAmount = selectedTokenAmount,
                         fiat = fiatCurrencyMode.toSellFiatCurrency()
                     )
                     onSellQuoteLoaded(sellQuote)
                 } catch (error: Throwable) {
                     handleError(error)
+                } finally {
+                    delay(SELL_QUOTE_REQUEST_DEBOUNCE_TIME)
```
1sakov, line 203: this will still have a delay if user will change the input, right? 
Why do we need 10 secs delay?

1sakov (approved):

PR #1185, opened 2023-01-30
PWN-6927 - switch to sol if there is no pools in calculations

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6927

## Description of Work

PWN-6927 - switch to sol if there is no pools in calculations


app/src/main/java/org/p2p/wallet/newsend/SendFeeRelayerManager.kt
```diff
@@ -131,14 +132,17 @@ class SendFeeRelayerManager(
...
                 sourceToken = sourceToken,
                 feeRelayerFee = feeRelayerFee,
                 feePayerToken = feePayer,
                 inputAmount = inputAmount,
                 strategy = strategy
             )
+        } catch (illegalStateException: IllegalStateException) {
+            Timber.i(illegalStateException, "Error during FeeRelayer fee calculation try to switch to sol")
+            sendInteractor.switchFeePayerToSol(solToken)
+            recalculate(sourceToken, inputAmount)
```
1sakov, line 145: I think it's a wrong place to handle that. 
There will be not only fee exceptions come here. 

You better catch it here: 

```
private suspend fun calculateFeeRelayerFee
```

gslevinkov (approved):

Davran331122 (approved):

PR #1189, opened 2023-01-31
NO-TICKET - fix icon padding

gslevinkov:


app/src/main/java/org/p2p/wallet/splash/SplashPresenter.kt
```diff
@@ -12,16 +11,13 @@ private const val MINIMUM_SPLASH_SHOWING_TIME_MS = 2000L
...
 class SplashPresenter(
     private val authInteractor: AuthInteractor,
-    private val sellInteractor: SellInteractor,
     private val userInteractor: UserInteractor
 ) : BasePresenter<SplashContract.View>(), SplashContract.Presenter {
 
     override fun attach(view: SplashContract.View) {
         super.attach(view)
         launch {
-            sellInteractor.loadSellAvailability()
```
gslevinkov, line 22: i think fetching ip flags in RootPresenter is okay too?

ui-kit/src/main/res/layout/widget_transaction_image.xml
```diff
@@ -13,7 +13,7 @@
         android:layout_height="match_parent"
         android:layout_margin="2dp"
         android:background="@drawable/bg_token_rounded"
-        android:padding="12dp"
+        app:contentPadding="12dp"
```
1sakov, line 16: Is it the only place which should be fixed?

1sakov (approved):

PR #1192, opened 2023-02-01
PWN-6887 - make empty state for history screens

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6887

## Description of Work

PWN-6887 - make empty state for history screens

## Screenshots (Optional)

Screenshot 1 | Screenshot 2 | Screenshot 3
[Android Emulator - Resizable_API_33:5554 2023-02-01 17-54-16] | [Android Emulator - Resizable_API_33:5554 2023-02-01 17-54-30] | [Paste Pic 3]


app/src/main/java/org/p2p/wallet/home/MainFragment.kt
```diff
@@ -65,6 +71,9 @@ class MainFragment : BaseFragment(R.layout.fragment_main), MainTabsSwitcher, Cen
 
     private var lastSelectedItemId = R.id.homeItem
 
+    private val mainScreenActionButtonHandler: MainScreenActionButtonHandler?
+        get() = tabCachedFragments[R.id.homeItem] as? MainScreenActionButtonHandler
```
1sakov, line 75: Why did you add a listener here, not directly in HomeFragment?

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/home/MainFragment.kt
kabramovp2p: according to what we have there is no way no do it
we need to call function from historyFragment to homeFragment 
this was fast way

another way is to use DI or re write code and make calls to buy and receive in MainFragment 

or add same code in history fragment

also we can use deeplinks in app

but all this consumes time and I think must be done in separate task

what do you think ?

app/src/main/java/org/p2p/wallet/history/ui/token/TokenHistoryFragment.kt
```diff
@@ -99,6 +99,8 @@ class TokenHistoryFragment :
         errorStateLayout.buttonRetry.setOnClickListener { presenter.retryLoad() }
         historyRecyclerView.setupHistoryList()
         viewActionButtons.onButtonClicked = { onActionButtonClicked(it) }
+        emptyStateLayout.buttonBuy.setOnClickListener { onActionButtonClicked(ActionButton.BUY_BUTTON) }
```
gslevinkov, line 102: what's wrong with opening just Buy and Recieve fragments here?

app/src/main/res/drawable/ic_history_empty.xml
```diff
@@ -0,0 +1,206 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
gslevinkov, line 1: that's a large fcking svg, have you seen the warning from AS?

`Very long vector path (1432 characters), which is bad for performance. Considering reducing precision, removing minor details or rasterizing vector`

kabramovp2p: on all big pictures with vectors we have this info message

app/src/main/res/layout/fragment_token_history.xml
```diff
@@ -5,17 +5,20 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:background="@color/backgroundPrimary"
+    android:background="@color/bg_smoke"
     android:orientation="vertical">
 
-    <androidx.appcompat.widget.Toolbar
+    <org.p2p.uikit.organisms.UiKitToolbar
```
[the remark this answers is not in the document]
kabramovp2p: move to common layout for reuse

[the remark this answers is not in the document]
gslevinkov: it doesn't solve the problem that we need to duplicate all the logic that fetches, filteres, builds the list on both screens

but thank you

gslevinkov (approved):

1sakov (approved):

PR #1199, opened 2023-02-02
PWN-7021 - swap quote

NikolaiEmelianov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7021

## Description of Work

swap quote request


app/src/main/java/org/p2p/wallet/swap/jupiter/api/SwapJupiterApi.kt
```diff
@@ -1,13 +1,27 @@
...
 
 interface SwapJupiterApi {
     @POST("v4/swap")
     suspend fun createRouteSwapTransaction(
         @Body body: CreateSwapTransactionRequest
     ): CreateSwapTransactionResponse
+
+    @GET("v4/quote")
+    suspend fun getQuote(
+        @Query("inputMint") inputMint: Base64String,
```
gslevinkov, line 22: i think mints are Base58String
also im not sure that retrofit will convert Base58String to String automatically if it's not @Body

app/src/main/java/org/p2p/wallet/swap/jupiter/api/response/quote/JupiterRouteResponse.kt
```diff
@@ -0,0 +1,27 @@
+package org.p2p.wallet.swap.jupiter.api.response.quote
+
+import com.google.gson.annotations.SerializedName
+import org.p2p.wallet.swap.jupiter.api.request.JupiterSwapFeesRequest
+import org.p2p.wallet.swap.jupiter.api.request.SwapRouteRequest
+import org.p2p.wallet.swap.jupiter.repository.model.JupiterSwapMode
+
+data class JupiterRouteResponse(
```
gslevinkov, line 8: maybe we can create just one entity and use it as request and response?

app/src/main/java/org/p2p/wallet/swap/jupiter/repository/JupiterSwapRoutesRepository.kt
```diff
@@ -1,3 +1,80 @@
...
+import kotlinx.coroutines.withContext
+
+class JupiterSwapRoutesRepository(
+    private val api: SwapJupiterApi,
+    private val dispatchers: CoroutineDispatchers,
+) : SwapRoutesRepository {
+
+    override suspend fun getSwapQuote(swapQuote: SwapQuote, userPublicKey: Base58String): List<SwapRoute> =
+        withContext(dispatchers.io) {
+            api.getQuote(swapQuote.inputMint, swapQuote.outputMint, swapQuote.amount, userPublicKey)
```
gslevinkov, line 24: named args

app/src/main/java/org/p2p/wallet/swap/jupiter/repository/JupiterSwapRoutesRepository.kt
```diff
@@ -1,3 +1,80 @@
...
+import org.p2p.wallet.utils.toBase58Instance
+import org.p2p.wallet.utils.toLamportsInstance
+import kotlinx.coroutines.withContext
+
+class JupiterSwapRoutesRepository(
+    private val api: SwapJupiterApi,
+    private val dispatchers: CoroutineDispatchers,
+) : SwapRoutesRepository {
+
+    override suspend fun getSwapQuote(swapQuote: SwapQuote, userPublicKey: Base58String): List<SwapRoute> =
```
gslevinkov, line 22: if method is named getSwapQuote - why i should pass SwapQuote here?

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/swap/jupiter/api/response/quote/JupiterRouteResponse.kt
NikolaiEmelianov: really its same

gslevinkov: yeah, that's my point, so instead of two same entities we can use one?

app/src/main/java/org/p2p/wallet/swap/jupiter/api/SwapJupiterApi.kt
```diff
@@ -18,9 +17,9 @@ interface SwapJupiterApi {
     ): CreateSwapTransactionResponse
 
     @GET("v4/quote")
-    suspend fun getQuote(
-        @Query("inputMint") inputMint: Base64String,
-        @Query("outputMint") outputMint: Base64String,
+    suspend fun getSwapRoutes(
+        @Query("inputMint") inputMint: Base58String,
```
gslevinkov, line 21: have checked that retrofit takes base58Value inside instead of calling Base58String.toString()?

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/swap/jupiter/api/SwapJupiterApi.kt
NikolaiEmelianov: org.p2p.wallet.home.model.Base58TypeAdapter
yes

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/utils/LamportsAmount.kt
```diff
@@ -8,3 +8,4 @@ data class LamportsAmount(val value: BigDecimal) {
 }
 
 fun BigDecimal.toLamportsInstance(): LamportsAmount = LamportsAmount(this)
+fun Double.toLamportsInstance(): LamportsAmount = toBigDecimal().toLamportsInstance()
```
1sakov, line 11: This is incorrect class. 
Lamports mean `BigInteger` 
What is the idea of this class? @gslevinkov

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/swap/jupiter/api/response/quote/JupiterRouteResponse.kt
1sakov: no, make it separate

NikolaiEmelianov: its same in swagger 
https://quote-api.jup.ag/v4/docs/static/index.html

app/src/main/java/org/p2p/wallet/utils/LamportsAmount.kt
gslevinkov: oh yeah, sorry, i made a mistake, the inside value should be BigInteger.
Jupiter return values in Lamports, but the type in API is double and I thought it would be big decimal.

the idea is use types instead of god know what like we do with Base58String and Base64String, we always know what format string has and compiler helps us. 
LamportsAmount - same idea. instead of adding `InLamports` everywhere - we can use type that tells us that this value / parameter - is in Lamports.

app/src/main/java/org/p2p/wallet/swap/jupiter/repository/JupiterSwapRoutesRepository.kt
```diff
@@ -1,3 +1,24 @@
...
-class JupiterSwapRoutesRepository
+import org.p2p.wallet.infrastructure.dispatchers.CoroutineDispatchers
+import org.p2p.wallet.swap.jupiter.api.SwapJupiterApi
+import org.p2p.wallet.swap.jupiter.repository.model.JupiterSwap
+import org.p2p.wallet.swap.jupiter.repository.model.SwapRoute
+import org.p2p.wallet.utils.Base58String
+
+class JupiterSwapRoutesRepository(
+    private val api: SwapJupiterApi,
+    private val dispatchers: CoroutineDispatchers,
```
1sakov, line 11: unused

1sakov (approved):

PR #1201, opened 2023-02-02
PWN-7066 - Try to fix blur mask effect

Davran331122:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7066


kabramovp2p (approved):

gslevinkov (approved):

app/src/main/AndroidManifest.xml
```diff
@@ -45,6 +45,7 @@
         <activity
             android:name="org.p2p.wallet.root.RootActivity"
             android:exported="true"
+            android:hardwareAccelerated="false"
```
NikolaiEmelianov, line 48: are you sure to disable it for whole app?

NikolaiEmelianov: https://developer.android.com/topic/performance/hardware-accel#view-level
maybe only for view?

NikolaiEmelianov (changes requested):

NikolaiEmelianov (approved):

PR #1218, opened 2023-02-07
PWN-6886 - fix error state to show on fetch history

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6886

## Description of Work

PWN-6886 - fix error state to show on fetch history


app/src/main/java/org/p2p/wallet/history/ui/history/HistoryPresenter.kt
```diff
@@ -102,7 +102,7 @@ class HistoryPresenter(
             val deferredBlockChainTransactions = async { fetchBlockChainTransactions(isRefresh) }
             moonpayTransactionsList = deferredSellTransactions.await()
             blockChainTransactionsList += deferredBlockChainTransactions.await()
-            if (moonpayTransactionsList.isFailed && blockChainTransactionsList.isFailed) {
+            if (moonpayTransactionsList.isFailed || blockChainTransactionsList.isFailed) {
```
gslevinkov, line 105: im not sure that's right
if we have moonpayTransactionList but failed blockhain list - we need to show moonpay transactions

gslevinkov (changes requested):

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryPresenter.kt
kabramovp2p: for now if you disable internet and go to history it will be empty screen not error
it is 146% issue

kabramovp2p: @gslevinkov if there is internet connection issue moonpayTransactionsList.isFailed == false
maybe here is an issue
can you fix it or you know where to fix?

gslevinkov: check the fetchMoonpayTransaction i guess, but this makes no sense
there will be absolute new logic for presenter, why fixing it now?

kabramovp2p: for QA to check error state is ok in https://p2pvalidator.atlassian.net/browse/PWN-6886
lol

gslevinkov: error state would be again not ok if you rewrite logic in presenter, Davran or you will just remove all the code here and write own logic
well, if you want to fix it now - okay, i won't stop you, i just asked

kabramovp2p: ok ok lets then just make a build for QA without merge

Davran331122 (approved):

kabramovp2p:
@gslevinkov fixed

gslevinkov (approved):

PR #1221, opened 2023-02-07
PWN-6935 - fix small usd amount

NikolaiEmelianov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6935

## Description of Work

fix small usd amount


Davran331122 (approved):

core/src/main/java/org/p2p/core/utils/AmountExtensions.kt
```diff
@@ -96,7 +96,12 @@ fun BigInteger.isZeroOrLess() = isZero() || isLessThan(BigInteger.ZERO)
 fun BigDecimal.asCurrency(currency: String): String =
     if (lessThenMinValue()) "<$currency 0.01" else "$currency ${formatFiat()}"
 
-fun BigDecimal.asUsd(): String = if (lessThenMinValue()) "<$ 0.01" else "$ ${formatFiat()}"
+fun BigDecimal.asUsd(tokenAmount: BigDecimal? = null): String =
```
1sakov, line 99: token cannot be null. 
Why did you add this check?

1sakov (changes requested):
comments

core/src/main/java/org/p2p/core/utils/AmountExtensions.kt
NikolaiEmelianov: Usd BigDecimal rounded early in lot of places. and in formatting time we have 
someToken = 0.00001
fiatAmount = 0.00 but it must be some 0.000005 depends of rate

1sakov: got it. 
The only then is to `tokenAmount` not nullable, right?

gslevinkov (approved):

Davran331122 (approved):

core/src/main/java/org/p2p/core/utils/AmountExtensions.kt
1sakov: I propose you this change: 

`fun BigDecimal.asUsd(excludeZerosCheck: Boolean = true): String = if (lessThenMinValue(excludeZerosCheck)) "<$ 0.01" else "$ ${formatFiat()}"`

`fun BigDecimal.lessThenMinValue(excludeZerosCheck: Boolean = true) = 
    if (excludeZerosCheck) {
         isLessThan(AMOUNT_MIN_VALUE.toBigDecimal())
     } else {
         isNotZero() &&  isLessThan(AMOUNT_MIN_VALUE.toBigDecimal())
     }
`

` fun getFormattedUsdTotal(): String? = totalInUsd?.asUsd(excludeZerosCheck= true)`

And you will work with amount then because `getFormattedUsdTotal` is available only for Token.Active where total is not zero

1sakov (changes requested):

core/src/main/java/org/p2p/core/utils/AmountExtensions.kt
NikolaiEmelianov: `totalInUsd` rounded in constructor. totalInUsd == 0.00 tokenAmount not
org/p2p/wallet/home/model/TokenConverter.kt:45

1sakov (approved):

PR #1251, opened 2023-02-15
PWN-6505 - Sentry optimization

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-6505


app/src/main/java/org/p2p/wallet/auth/interactor/UsernameInteractor.kt
```diff
@@ -57,7 +57,7 @@ class UsernameInteractor(
             }
             crashLogger.setCustomKey("username", usernameDetails?.username?.fullUsername.orEmpty())
         } catch (error: Throwable) {
-            Timber.i(error, "Failed to restore username for ${owner.base58Value}")
+            Timber.e(error, "Failed to restore username for ${owner.base58Value}")
```
1sakov, line 60: remove `owner.base58Value`

app/src/main/java/org/p2p/wallet/infrastructure/network/provider/SeedPhraseProvider.kt
```diff
@@ -30,7 +30,7 @@ class SeedPhraseProvider(
         set(value) {
             field = value
             saveSeedPhraseToStorage(value)
-            Timber.tag(TAG).i("updating user seed phrase: $value")
+            Timber.tag(TAG).i("updating user seed phrase: ${value.size}")
```
1sakov, line 33: remove `value.size`

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/auth/interactor/UsernameInteractor.kt
gslevinkov: why? we need to know which user is failed to restore username

app/src/main/java/org/p2p/wallet/infrastructure/network/provider/SeedPhraseProvider.kt
gslevinkov: replaced with isEmpty to understand are we removing seedphrase from shared prefs

on reset we are doing seedPhrase = emptyList()

app/src/main/java/org/p2p/wallet/auth/interactor/UsernameInteractor.kt
1sakov: someone can easily pass an owner as a secret key to this function. 
we will know which user is failed from anywhere

gslevinkov: ok, removing

1sakov (approved):

PR #1258, opened 2023-02-15
PWN-7063 - deeplink onNewIntent Fix

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7063

## Description of Work

PWN-7063 - if onNewIntentCalled with fast onboarding deeplink make replace on splash fragment to use same deeplink logic


app/src/main/java/org/p2p/wallet/root/RootActivity.kt
```diff
@@ -14,6 +14,9 @@ import com.google.android.gms.common.ConnectionResult
 import com.google.android.gms.common.GoogleApiAvailability
 import com.google.android.material.snackbar.Snackbar
 import org.koin.android.ext.android.inject
+import timber.log.Timber
```
kabramovp2p, line 17: image

XZ

PR #1271, opened 2023-02-17
PWN-5812 - Change history environment config, to push service

Davran331122:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5812

### DIFF
--- a/app/src/main/java/org/p2p/wallet/infrastructure/network/interceptor/RpcInterceptor.kt
+++ b/app/src/main/java/org/p2p/wallet/infrastructure/network/interceptor/RpcInterceptor.kt
@@ -7,7 +7,6 @@ import okhttp3.Interceptor
 import okhttp3.Request
 import okhttp3.Response
 import okhttp3.ResponseBody.Companion.toResponseBody
-import okio.Buffer
 import org.json.JSONArray
 import org.json.JSONException
 import org.json.JSONObject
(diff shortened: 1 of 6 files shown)


thebrodyaga (approved):

app/src/main/java/org/p2p/wallet/infrastructure/network/interceptor/RpcInterceptor.kt
```diff
@@ -46,50 +44,12 @@ class RpcInterceptor(
...
-        val json = getRequestJson(request)
         val url = request.url
-        val networkEnvironmentForRequest = getValidatedNetworkEnvironment(json)
         return request.newBuilder()
-            .url(createRpcUrl(url, networkEnvironmentForRequest))
+            .url(createRpcUrl(url, currentEnvironment))
             .build()
     }
 
-    private fun getRequestJson(request: Request): JSONObject? {
```
gslevinkov, line 58: why have you removed this one?

NikolaiEmelianov (approved):

app/src/main/java/org/p2p/wallet/infrastructure/network/interceptor/RpcInterceptor.kt
Davran331122: it's have not usage

Davran331122: it was used to determine environment for history requests, for now we don't use this requests, so there is no need in this function

kabramovp2p (approved):

PR #1281, opened 2023-02-20
PWN-5813 - added fee parsing for details to show

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5813

## Description of Work

PWN-5813 - added fee parsing for details to show


app/src/main/java/org/p2p/wallet/history/interactor/mapper/RpcHistoryTransactionConverter.kt
```diff
@@ -258,3 +270,22 @@ private fun RpcHistoryTypeResponse.toDomain(): RpcHistoryTransactionType {
...
 }
+
+private fun List<RpcHistoryFeeResponse>.formatFees(lessThenMinString: String): String? {
+    return if (this.all { fee -> fee.payer == Constants.FEE_RELAYER_ACCOUNT }) {
+        null
+    } else {
+        return joinToString(separator = "\n") { fee ->
+            val feeInFiat = fee.amount?.usdAmount.toBigDecimalOrZero()
+            val formattedUsdAmount = if (feeInFiat.lessThenMinValue()) lessThenMinString
+            else "$${feeInFiat.formatFiat()}"
```
gslevinkov, line 281: brackets

app/src/main/java/org/p2p/wallet/history/interactor/mapper/RpcHistoryTransactionConverter.kt
```diff
@@ -258,3 +270,22 @@ private fun RpcHistoryTypeResponse.toDomain(): RpcHistoryTransactionType {
         RpcHistoryTypeResponse.UNKNOWN -> RpcHistoryTransactionType.UNKNOWN
     }
 }
+
+private fun List<RpcHistoryFeeResponse>.formatFees(lessThenMinString: String): String? {
+    return if (this.all { fee -> fee.payer == Constants.FEE_RELAYER_ACCOUNT }) {
+        null
+    } else {
+        return joinToString(separator = "\n") { fee ->
```
gslevinkov, line 278: no need for return

gslevinkov (approved):

core/src/main/java/org/p2p/core/utils/Constants.kt
```diff
@@ -24,6 +24,7 @@ object Constants {
     const val USDT_COINGECKO_ID = "tether"
 
     const val WRAPPED_SOL_MINT = "So11111111111111111111111111111111111111112"
+    const val FEE_RELAYER_ACCOUNT = "FG4Y3yX4AAchp1HvNZ7LfzFTewF2f6nDoMDCohTFrdpT"
```
1sakov, line 27: Don't we have some value in the response to determine that payer is fee relayer account?
It may be changed and you'll have incorrect fees

1sakov (approved):

core/src/main/java/org/p2p/core/utils/Constants.kt
kabramovp2p: we do not
and i thought that it is our feeRelayer address and it is a constant

PR #1305, opened 2023-02-23
PWN-7169 - back pressed, check main tab

NikolaiEmelianov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7169

## Description of Work

back pressed, check main tab

### DIFF
--- a/app/src/main/java/org/p2p/wallet/swap/ui/jupiter/main/JupiterSwapFragment.kt
+++ b/app/src/main/java/org/p2p/wallet/swap/ui/jupiter/main/JupiterSwapFragment.kt
@@ -1,5 +1,6 @@
 package org.p2p.wallet.swap.ui.jupiter.main
 
+import androidx.activity.addCallback
 import androidx.core.view.isInvisible
 import android.os.Bundle
 import android.view.View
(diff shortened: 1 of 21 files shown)


app/src/main/java/org/p2p/wallet/swap/ui/jupiter/main/JupiterSwapFragment.kt
```diff
@@ -61,6 +64,14 @@ class JupiterSwapFragment :
             imageViewSwapTokens.setOnClickListener {
                 presenter.switchTokens()
             }
+            val onBackPressed = { presenter.onBackPressed() }
+            toolbar.setNavigationOnClickListener { onBackPressed() }
+            if (isInMainTabsScreen()) {
```
gslevinkov, line 69: i personally handle this cases with enum as an arg and checking here with if

it eliminates additional fragment logic and interfaces that you created

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/swap/ui/jupiter/main/JupiterSwapFragment.kt
NikolaiEmelianov: imho easy to use extension than to create some extra_arg, extra_arg_key, add new method params, set param in other features or when parse deeplink

kabramovp2p (approved):

PR #1326, opened 2023-02-27
PWN-5812 - Fix empty state

Davran331122:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-5812

### DIFF
--- a/app/src/main/java/org/p2p/wallet/history/ui/historylist/HistoryListView.kt
+++ b/app/src/main/java/org/p2p/wallet/history/ui/historylist/HistoryListView.kt
@@ -64,7 +64,6 @@ class HistoryListView @JvmOverloads constructor(
     }
 
     private fun bindView() {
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/p2p/wallet/history/ui/historylist/HistoryListView.kt
```diff
@@ -64,7 +64,6 @@ class HistoryListView @JvmOverloads constructor(
     }
 
     private fun bindView() {
-        presenter.attach(this)
```
1sakov, line 67: why?

Davran331122: it's in init block

1sakov (approved):

kabramovp2p (approved):

PR #1336, opened 2023-03-01
PWN-7116 - select routes fragment

NikolaiEmelianov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7116

## Description of Work

select routes fragment


app/src/main/java/org/p2p/wallet/swap/ui/jupiter/routes/SwapSelectRoutesBottomSheet.kt
```diff
@@ -0,0 +1,95 @@
...
+    private val adapter = SwapSettingsAdapter(
+        baseCellDelegate(inflateListener = { financeBlock ->
+            financeBlock.setOnClickAction { view, item -> onRouteClick(item, view) }
+        }),
+    )
+
+    override fun getTheme(): Int = R.style.WalletTheme_BottomSheet_Rounded
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+        return inflater.inflate(R.layout.dialog_swap_select_routes, container, false)
```
gslevinkov, line 63: i think by viewBinding already inflates the layout, no?

app/src/main/java/org/p2p/wallet/swap/ui/jupiter/routes/SwapSelectRoutesMapper.kt
```diff
@@ -0,0 +1,81 @@
...
+                    )
+                ),
+                rightSideCellModel = rightIcon,
+                payload = route,
+            )
+        }
+    }
+
+    private fun formatLabel(route: JupiterSwapRoute): String {
+        var result = ""
```
gslevinkov, line 56: better to use buildString instead of raw .plus

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/swap/ui/jupiter/routes/SwapSelectRoutesBottomSheet.kt
1sakov: no, we need to pass it

1sakov (approved):

PR #1340, opened 2023-03-02
PWN-7174 - New Swap. Bottom padding and popular tokens

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7174


kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/swap/ui/jupiter/tokens/SwapTokensFragment.kt
```diff
@@ -63,11 +65,17 @@ class SwapTokensFragment :
         binding.toolbar.setNavigationOnClickListener { popBackStack() }
         inflateSearchMenu(binding.toolbar)
 
-        binding.recyclerViewTokens.attachAdapter(adapter)
-        when (tokenToChange) {
-            SwapTokensListMode.TOKEN_A -> binding.recyclerViewTokens.addItemDecoration(SwapTokenAItemDecoration())
-            SwapTokensListMode.TOKEN_B -> {
-                // todo
+        with(binding.recyclerViewTokens) {
+            attachAdapter(this@SwapTokensFragment.adapter)
```
1sakov, line 69: Where is layout manager? 
Please set it in code, it's not obvious if it's in xml.
If we need to modify it, we will need to remove it from xml and use in code anyway

1sakov (approved):
Make sure, your unit tests will not fail after PR is merged

PR #1344, opened 2023-03-02
PWN-7379 - show claim info bottomSheet with claim details

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7379

## Description of Work

PWN-7379 - show claim info bottomSheet with claim details

## Screenshot

Screenshot_20230302_180529


app/src/main/java/org/p2p/wallet/claim/model/ClaimFee.kt
```diff
@@ -0,0 +1,31 @@
...
+import android.os.Parcelable
+import java.math.BigDecimal
+import kotlinx.parcelize.Parcelize
+import org.p2p.core.utils.asApproximateUsd
+import org.p2p.core.utils.formatToken
+import org.p2p.core.utils.scaleMedium
+import org.p2p.wallet.utils.emptyString
+
+@Parcelize
+data class ClaimFee(
```
1sakov, line 12: `BridgeFee`?

app/src/main/java/org/p2p/wallet/claim/model/ClaimDetails.kt
```diff
@@ -0,0 +1,12 @@
+package org.p2p.wallet.claim.model
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+data class ClaimDetails(
+    val willGet: ClaimFee,
```
1sakov, line 8: `amount`
And it's not `ClaimFee` right?

app/src/main/java/org/p2p/wallet/claim/ui/dialogs/ClaimInfoBottomSheet.kt
```diff
@@ -0,0 +1,88 @@
...
+    companion object {
+        fun show(
+            fm: FragmentManager,
+            claimDetails: ClaimDetails
+        ) {
+            ClaimInfoBottomSheet().withArgs(
+                ARG_CLAIM_DETAILS to claimDetails
+            ).show(
+                fm, ClaimInfoBottomSheet::javaClass.name
+            )
```
1sakov, line 32:
```suggestion
            ClaimInfoBottomSheet()
                  .withArgs(ARG_CLAIM_DETAILS to claimDetails)
                  .show(fm, ClaimInfoBottomSheet::javaClass.name)
```

app/src/main/res/layout/dialog_claim_info.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:orientation="vertical">
+
+    <ImageView
+        android:layout_width="220dp"
+        android:layout_height="164dp"
```
1sakov, line 10: why so? 
Make it wrap content

app/src/main/res/values/bridges_strings.xml
```diff
@@ -11,4 +11,11 @@
     <string name="claiming_fees_free">Paid by Key App</string>
     <string name="claiming_bottom_button">Send</string>
 
+    <string name="claim_info_free_transaction_message">With Key App, the first transaction is free. Also all the transactions above $300 are free</string>
+    <string name="claim_info_transaction_free">Free</string>
+    <string name="claim_info_you_will_get">You will get</string>
+    <string name="claim_info_network_fee">Network Fee</string>
+    <string name="claim_info_account_creation_fee">Account creation Fee</string>
+    <string name="claim_info_bridge_fee">Wormhole Bridge and Transaction Fee</string>
```
1sakov, line 19: I would name it all as `bridge_some_key`, not claim.
Because claim is some flow of the whole bridge feature

1sakov (approved):

app/src/main/java/org/p2p/wallet/claim/model/ClaimDetails.kt
kabramovp2p: heh it is fee and amount
i think i need to rename it in common way

app/src/main/res/values/bridges_strings.xml
kabramovp2p: yes claim _ send _ receive
i thought to name it according to the flow but ok will rename it

app/src/main/res/layout/dialog_claim_info.xml
kabramovp2p: because this image is smaller then we need
=( it is from sell but we can not have duplicates

app/src/main/java/org/p2p/wallet/claim/ui/dialogs/ClaimInfoBottomSheet.kt
```diff
@@ -0,0 +1,88 @@
...
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        setDoneButtonVisibility(isVisible = false)
+        with(binding) {
+            layoutFreeTransactions.bindLayoutFreeTransactions()
+            layoutWillGetAmount.bindDetailsLineWithFee(
+                title = getString(R.string.claim_info_you_will_get),
+                fee = claimDetails.willGet
+            )
+            layoutNetworkFee.bindDetailsLineWithFee(
```
gslevinkov, line 54: looks nice, thanks!

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/claim/ui/dialogs/ClaimInfoBottomSheet.kt
```diff
@@ -0,0 +1,88 @@
...
+        ) {
+            ClaimInfoBottomSheet().withArgs(
+                ARG_CLAIM_DETAILS to claimDetails
+            ).show(
+                fm, ClaimInfoBottomSheet::javaClass.name
+            )
+        }
+    }
+
+    private lateinit var binding: DialogClaimInfoBinding
```
gslevinkov, line 36: override fun onDestroyView() {
    super.onDestroyView()
    _binding = null
}

gslevinkov: https://developer.android.com/topic/libraries/view-binding

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/claim/ui/dialogs/ClaimInfoBottomSheet.kt
kabramovp2p: wow wow
GC does not work for that?
we did not do something like that ever

PR #1345, opened 2023-03-03
PWN-7176 - New swap. Slippage error design

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7176

Screen Shot 2023-03-03 at 15 29 32


app/src/main/java/org/p2p/wallet/transaction/ui/JupiterTransactionProgressBottomSheet.kt
```diff
@@ -130,49 +133,90 @@ class JupiterTransactionProgressBottomSheet : BottomSheetDialogFragment() {
...
+        dismissResult = JupiterTransactionDismissResult.TrySwapAgain
+
+        textViewTitle.text = progressStateFormat.format(getString(R.string.transaction_progress_failed))
+        progressStateTransaction.setFailedState()
+        progressStateTransaction.setDescriptionText(R.string.swap_transaction_details_error_unknown)
+        buttonDone.setText(R.string.common_try_again)
+    }
+
+    private fun getNewSlippage(currentSlippage: Double): Double = when (currentSlippage) {
+        in (0.0..0.4) -> 0.5
```
Davran331122, line 208: can we move it so fun or to const or to class?

Davran331122 (approved):

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/transaction/ui/JupiterTransactionProgressBottomSheet.kt
gslevinkov: i think i can, but i cant think of any good name for this const
it's very easy mapping from Figma, looks good i think

NikolaiEmelianov (approved):

PR #1348, opened 2023-03-03
PWN-7452 - Swap rate ticker added

1sakov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7452

## Description of Work

## Screenshots (Optional)

### DIFF
--- a/app/src/main/java/org/p2p/wallet/swap/jupiter/interactor/model/SwapTokenModel.kt
+++ b/app/src/main/java/org/p2p/wallet/swap/jupiter/interactor/model/SwapTokenModel.kt
@@ -2,6 +2,8 @@ package org.p2p.wallet.swap.jupiter.interactor.model
 
 import java.math.BigDecimal
 import org.p2p.core.token.Token
+import org.p2p.core.utils.Constants.USDC_SYMBOL
+import org.p2p.core.utils.Constants.USDT_SYMBOL
 import org.p2p.wallet.swap.jupiter.repository.model.JupiterSwapToken
 import org.p2p.wallet.utils.Base58String
 import org.p2p.wallet.utils.toBase58Instance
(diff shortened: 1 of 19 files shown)


app/src/main/java/org/p2p/wallet/swap/jupiter/interactor/model/SwapTokenModel.kt
```diff
@@ -14,6 +16,8 @@ sealed interface SwapTokenModel {
     val tokenSymbol: String
     val iconUrl: String?
 
+    fun isStable(): Boolean = tokenSymbol == USDC_SYMBOL || tokenSymbol == USDT_SYMBOL
```
gslevinkov, line 19: maybe isStableCoin? 
we can see model.isStable() and don't understand without looking inside

app/src/main/java/org/p2p/wallet/swap/jupiter/statemanager/rate/SwapRateTickerManager.kt
```diff
@@ -0,0 +1,133 @@
...
+
+    fun onInitialTokensSelected(
+        newTokenA: SwapTokenModel?,
+        newTokenB: SwapTokenModel?
+    ) {
+        if (newTokenA == null || newTokenB == null) return
+
+        val currentTokenSymbolA = currentTokenA?.tokenSymbol
+        val currentTokenSymbolB = currentTokenB?.tokenSymbol
+        if (newTokenA.tokenSymbol == currentTokenSymbolA && newTokenB.tokenSymbol == currentTokenSymbolB) {
```
gslevinkov, line 78: should we ignore CASE for symbols when using equals? 
because SwapTokenModel can be UserToken (Token.Active from solanalist.json) and JupiterToken (from jupiter API). maybe there is a change that they will have different naming. 
maybe even better - check for mint addresses instead of symbols

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/swap/jupiter/statemanager/rate/SwapRateTickerManager.kt
1sakov: sounds good, will do

app/src/main/java/org/p2p/wallet/swap/jupiter/interactor/model/SwapTokenModel.kt
1sakov: ok

PR #1350, opened 2023-03-03
PWN-7229 - New swap. Save selected tokens on change

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7229


core/src/main/java/org/p2p/core/token/Token.kt
```diff
@@ -188,3 +188,7 @@ sealed class Token constructor(
         }
     }
 }
+
+fun List<Token.Active>.findSol(): Token.Active = first { it.isSOL }
```
1sakov, line 192: `requireSol` would be better to show that it can throw exception

1sakov (approved):

core/src/main/java/org/p2p/core/token/Token.kt
gslevinkov: good idea! but it's just doesn't look semantic right

tokens.requireSol() looks like just a check
i made a contract based on List operations like firstOrNull first() that we are all know

require, requireNotNull from kotlin checks have return type `Nothing`

1sakov: Okay. 
You may add `throw` maybe, smth like `findSolOrThrow` but it's very optional.

PR #1376, opened 2023-03-07
PWN-7526 - default item decoration

NikolaiEmelianov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7526

## Description of Work

default item decoration

### DIFF
--- a/ui-kit/src/main/kotlin/org/p2p/uikit/utils/recycler/decoration/EightPxOffsetDecoration.kt
+++ b/ui-kit/src/main/kotlin/org/p2p/uikit/utils/recycler/decoration/EightPxOffsetDecoration.kt
@@ -1,0 +1,33 @@
+package org.p2p.uikit.utils.recycler.decoration
+
+import androidx.recyclerview.widget.RecyclerView
+import androidx.recyclerview.widget.RecyclerView.ItemDecoration
+import android.graphics.Rect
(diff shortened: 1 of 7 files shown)


ui-kit/src/main/kotlin/org/p2p/uikit/utils/recycler/decoration/EightPxOffsetDecoration.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.p2p.uikit.utils.recycler.getItems
+import org.p2p.uikit.utils.toPx
+
+fun eightOffsetFinanceBlockDecoration() = EightPxOffsetDecoration(FinanceBlockCellModel::class)
+
+class EightPxOffsetDecoration(
+    private val itemCellType: KClass<out AnyCellItem>
+) : ItemDecoration() {
+
+    private val offset = 8.toPx()
```
kabramovp2p, line 19: maybe we need to make an ability to custom offset
pass it to constructor like 

class EightPxOffsetDecoration(
    private val itemCellType: KClass<out AnyCellItem>,
    private val offsetDp = 8.toPx()
) : ItemDecoration() {

1sakov (approved):

gslevinkov (approved):

PR #1379, opened 2023-03-08
PWN-7533 - New Swap. Proxy backend

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7533
https://p2pvalidator.atlassian.net/browse/PWN-7541

## Description of Work
- removed recent_blockhash from SDK request

### DIFF
--- a/app/src/main/java/org/p2p/wallet/swap/jupiter/interactor/JupiterSwapInteractor.kt
+++ b/app/src/main/java/org/p2p/wallet/swap/jupiter/interactor/JupiterSwapInteractor.kt
@@ -7,7 +7,6 @@ import org.p2p.solanaj.model.types.Encoding
 import org.p2p.solanaj.rpc.RpcSolanaRepository
 import org.p2p.wallet.infrastructure.network.data.ServerException
 import org.p2p.wallet.infrastructure.network.provider.TokenKeyProvider
-import org.p2p.wallet.rpc.repository.blockhash.RpcBlockhashRepository
 import org.p2p.wallet.sdk.facade.RelaySdkFacade
 import org.p2p.wallet.swap.jupiter.interactor.model.SwapTokenModel
 import org.p2p.wallet.swap.jupiter.repository.model.JupiterSwapRoute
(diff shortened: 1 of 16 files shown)


1sakov (approved):

app/src/main/res/values/environment_config.xml
```diff
@@ -26,7 +26,7 @@
 
     <string name="orcaApiBaseUrl">https://orca.key.app/</string>
 
-    <string name="jupiterQuoteBaseUrl">https://quote-api.jup.ag/</string>
+    <string name="jupiterQuoteBaseUrl">https://swap.keyapp.org/</string>
```
1sakov, line 29: what about the prod url? 
When we'll add it?

app/src/main/java/org/p2p/wallet/swap/jupiter/interactor/JupiterSwapInteractor.kt
```diff
@@ -37,13 +35,14 @@ class JupiterSwapInteractor(
...
-        val latestBlockhash = rpcBlockhashRepository.getRecentBlockhash()
         val userAccount = Account(tokenKeyProvider.keyPair)
 
         val signedSwapTransaction = relaySdkFacade.signTransaction(
             transaction = swapTransaction,
             keyPair = userAccount.getEncodedKeyPair().toBase58Instance(),
-            recentBlockhash = latestBlockhash
+            // empty string because swap transaction already has recent blockhash
+            // if pass our own recent blockhash, there is an error
+            recentBlockhash = null
```
1sakov, line 45: but you are cleaning the current blockhash by this, no?

app/src/main/res/values/environment_config.xml
gslevinkov: there no prod ulr, we'll discuss it today

app/src/main/java/org/p2p/wallet/swap/jupiter/interactor/JupiterSwapInteractor.kt
gslevinkov: no, the current blockhash in transaction already exists from Jupiter, and our SDK overwrites it that causes error from RPC

kabramovp2p (approved):

PR #1398, opened 2023-03-09
NO-TICKET - Optimized the tokens data loading from solana.tokenlist.json

1sakov:
Now we are not loading tokens data every time when the app launches to download tokens data from
`https://raw.githubusercontent.com/p2p-org/token-list/main/src/tokens/solana.tokenlist.json`

Now we this data is stored in a separate file which is accessible and readable only from our app. 
There is `TokensDataWorker` added which is being launched every 3 days by WorkManager which updates this file


app/src/main/java/org/p2p/wallet/App.kt
```diff
@@ -78,6 +96,10 @@ class App : Application() {
         AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_FOLLOW_SYSTEM)
     }
 
+    private fun setupWorkManager() {
+        WorkManager.initialize(this, Configuration.Builder().build())
```
gslevinkov, line 100: Configuration.Builder().build() -> getWorkManagerConfiguration()

app/src/main/java/org/p2p/wallet/common/storage/ExternalStorageRepository.kt
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    fun readJsonFile(fileName: String): ExternalFile? {
+        return try {
+            context.openFileInput(fileName).bufferedReader().useLines { lines ->
+                val lastModified = File(context.filesDir, fileName).lastModified()
+                ExternalFile(lines.joinToString(""), lastModified)
+            }
+        } catch (e: Exception) {
+            e.printStackTrace()
```
gslevinkov, line 27: i think Timber.e is better

app/src/main/java/org/p2p/wallet/common/storage/ExternalStorageRepository.kt
```diff
@@ -0,0 +1,31 @@
...
+    private val context: Context
+) {
+
+    fun saveJson(json: String, fileName: String) {
+        try {
+            context.openFileOutput(fileName, Context.MODE_PRIVATE).use {
+                it.write(json.toByteArray())
+            }
+        } catch (e: Exception) {
+            e.printStackTrace()
```
gslevinkov, line 16: Timber

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/App.kt
1sakov: good catch, thank you

app/src/main/java/org/p2p/wallet/user/interactor/UserInteractor.kt
```diff
@@ -64,8 +72,26 @@ class UserInteractor(
...
+            if (tokens != null) {
+                Timber.tag(TAG).d("Tokens data were successfully parsed from file.")
+                userLocalRepository.setTokenData(tokens)
+                return
+            }
+        }
+
+        Timber.tag(TAG).d("Tokens data file was not found. Loading from remote")
+        // If the file is not found or empty, load from network
         val data = userRepository.loadAllTokens()
```
kabramovp2p, line 89: loading and saving same as in worker, we can make this code common I think

kabramovp2p (approved):

PR #1401, opened 2023-03-09
PWN-7380 - Optimize token list

Davran331122:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7380

### DIFF
--- a/app/src/main/java/org/p2p/wallet/receive/tokenselect/ReceiveTokensFragment.kt
+++ b/app/src/main/java/org/p2p/wallet/receive/tokenselect/ReceiveTokensFragment.kt
@@ -2,6 +2,8 @@ package org.p2p.wallet.receive.tokenselect
 
 import androidx.appcompat.widget.SearchView
 import androidx.appcompat.widget.Toolbar
+import androidx.core.view.doOnAttach
+import androidx.core.view.doOnDetach
 import androidx.core.view.isVisible
 import androidx.recyclerview.widget.DiffUtil
 import androidx.recyclerview.widget.LinearLayoutManager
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/p2p/wallet/receive/tokenselect/ReceiveTokensPresenter.kt
```diff
@@ -40,22 +53,31 @@ class ReceiveTokensPresenter(
...
         }
     }
 
     override fun onSearchTokenQueryChanged(newQuery: String) {
-        searchText = newQuery
-        if (newQuery.isBlank()) {
-            view?.resetScrollPosition()
+        searchJob?.cancel()
+        searchJob = launch {
+            delay(300L)
```
kabramovp2p, line 73: Noooo :( not delay :(

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/receive/tokenselect/ReceiveTokensFragment.kt
```diff
@@ -70,34 +72,40 @@ class ReceiveTokensFragment :
...
-                loadNextPage = { presenter.load(isRefresh = false) }
-            ).also {
-                addOnScrollListener(it)
+            doOnAttach {
+                layoutManager = this@ReceiveTokensFragment.layoutManager
+                addOnScrollListener(scrollListener)
+            }
+            doOnDetach {
+                layoutManager = null
+                removeOnScrollListener(scrollListener.also { it.reset() })
```
1sakov, line 100: awesome, thanks for fixing leaks

app/src/main/java/org/p2p/wallet/receive/tokenselect/ReceiveTokensPresenter.kt
1sakov: delay here is totally good, just move it to const please

1sakov (approved):

PR #1406, opened 2023-03-10
PWN-7571 - update supported tokens and use them in receive + add as a first tokens in receive

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7571

## Description of Work

PWN-7571 - update supported tokens and use them in receive + add as a first tokens in receive

### DIFF
--- a/app/src/main/java/org/p2p/wallet/receive/tokenselect/ReceiveTokensPresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/receive/tokenselect/ReceiveTokensPresenter.kt
@@ -37,14 +37,14 @@ class ReceiveTokensPresenter(
     private var searchJob: Job? = null
 
     private val tokensFlow = MutableStateFlow<List<AnyCellItem>>(emptyList())
-    private lateinit var pinedWormholeTokens: List<TokenData>
-    private lateinit var pinedWormholeTokensAddresses: List<String>
+    private lateinit var pinnedWormholeTokens: List<TokenData>
+    private lateinit var pinnedWormholeTokensAddresses: List<String>
 
     override fun attach(view: ReceiveTokensContract.View) {
         super.attach(view)
         launch {
-            pinedWormholeTokens = preparePinedWormholeTokens()
-            pinedWormholeTokensAddresses = pinedWormholeTokens.map { it.mintAddress }
+            pinnedWormholeTokens = preparePinedWormholeTokens()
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/p2p/wallet/receive/tokenselect/ReceiveTokensPresenter.kt
```diff
@@ -36,10 +37,14 @@ class ReceiveTokensPresenter(
     private var searchJob: Job? = null
 
     private val tokensFlow = MutableStateFlow<List<AnyCellItem>>(emptyList())
+    private lateinit var pinedWormholeTokens: List<TokenData>
+    private lateinit var pinedWormholeTokensAddresses: List<String>
```
1sakov, line 41: `pinned`

app/src/main/java/org/p2p/wallet/receive/tokenselect/ReceiveTokensPresenter.kt
```diff
@@ -57,6 +62,24 @@ class ReceiveTokensPresenter(
...
+                    mintAddress = token.mintAddress,
+                    name = erc20Token.replaceTokenName ?: token.tokenName,
+                    symbol = erc20Token.replaceTokenSymbol ?: token.tokenSymbol,
+                    iconUrl = erc20Token.tokenIconUrl ?: token.iconUrl,
+                    decimals = token.decimals,
+                    coingeckoId = erc20Token.coingeckoId,
+                    isWrapped = false,
+                    serumV3Usdc = null,
+                    serumV3Usdt = null
+                )
```
1sakov, line 78: can we do this mapping inside the interactor?
Do we have `EthereumInteractor`?

1sakov (approved):

app/src/main/java/org/p2p/wallet/receive/tokenselect/ReceiveTokensPresenter.kt
kabramovp2p: it is in Ethereum package and here we need TokenData that we can not access from EthereumInteractor

1sakov: looks like we need to move `TokenData` to core as well. 
But it's optional

NikolaiEmelianov (approved):

PR #1410, opened 2023-03-10
PWN-7176 - New swap. Correct slippage error detect

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7176

## Description of Work

Used:
https://docs.rs/solana-sdk/latest/solana_sdk/transaction/enum.TransactionError.html
https://github.com/jup-ag/jupiter-cpi/blob/main/idl.json


app/src/main/java/org/p2p/wallet/infrastructure/network/data/ServerException.kt
```diff
@@ -8,7 +8,8 @@ class ServerException(
     val errorCode: ErrorCode,
     fullMessage: String,
     private val errorMessage: String?,
-    val jsonErrorBody: JsonObject? = null
+    val jsonErrorBody: JsonObject? = null,
+    val domainErrorType: RpcTransactionError? = null
```
gslevinkov, line 12: doesn't break work with this class
still exception, but with additional nullable trycatched field

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/infrastructure/network/data/InstructionErrorType.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import com.google.gson.JsonElement
+
+private const val KEY_CUSTOM_INSTRUCTION_ERROR_ID = "Custom"
+
+sealed interface InstructionErrorType {
+    companion object {
+        // can be { "Custom" : 1111 } or just plain "SomeError"
+        fun from(instructionErrorType: JsonElement): InstructionErrorType {
+            val isCustom = instructionErrorType.isJsonPrimitive
```
Davran331122, line 11: Why if it's primitive we cast it as JsonObject ? if I am not wrong, String is primitive too.

app/src/main/java/org/p2p/wallet/infrastructure/network/data/RpcTransactionError.kt
```diff
@@ -0,0 +1,64 @@
...
+ * }
+ * ```
+ * @return valid typed RpcTransactionError or null if parsing is failed
+ */
+sealed interface RpcTransactionError {
+    companion object {
+        fun from(errJson: JsonObject): RpcTransactionError? {
+            // for example "InstructionError"
+            val transactionErrorType: String? = errJson.keySet().firstOrNull()
+            return when (transactionErrorType) {
```
Davran331122, line 32: you can wrap it to some enum class

app/src/main/java/org/p2p/wallet/infrastructure/network/data/RpcTransactionError.kt
```diff
@@ -0,0 +1,64 @@
...
+                "AccountInUse" -> AccountInUse
+                "InstructionError" -> parseInstructionError(errJson, transactionErrorType)
+                else -> null
+            }
+        }
+    }
+
+    object AccountInUse : RpcTransactionError
```
Davran331122, line 40: can AccountInUse and InstructionError be inherited from RpcTransactionError sealed class ?

app/src/main/java/org/p2p/wallet/jupiter/interactor/JupiterSwapInteractor.kt
```diff
@@ -47,9 +46,10 @@ class JupiterSwapInteractor(
         Timber.i("Swap tokens success: $firstTransactionSignature")
         JupiterSwapTokensResult.Success
     } catch (error: ServerException) {
-        val errorMessage = error.message.orEmpty()
-        if (LOW_SLIPPAGE_ERROR_CODE in errorMessage ||
-            LOW_SLIPPAGE_ERROR_MESSAGE in errorMessage
+        if (error.domainErrorType != null &&
```
Davran331122, line 49: you can move error.domainErrorType to variable

NikolaiEmelianov (approved):

app/src/main/java/org/p2p/wallet/infrastructure/network/data/InstructionErrorType.kt
gslevinkov: thanks for pointing this out, made a test to check this stuff

app/src/main/java/org/p2p/wallet/infrastructure/network/data/RpcTransactionError.kt
gslevinkov: no need for that, we know all the string values

app/src/main/java/org/p2p/wallet/infrastructure/network/data/RpcTransactionError.kt
gslevinkov: already inherited, or i haven't catched your question

app/src/main/java/org/p2p/wallet/jupiter/interactor/JupiterSwapInteractor.kt
gslevinkov: done

PR #1413, opened 2023-03-12
PWN-7382 - Fix receive ethereum ui bugs

Davran331122:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7382

### DIFF
--- a/app/src/main/java/org/p2p/wallet/receive/eth/ReceiveEthereumFragment.kt
+++ b/app/src/main/java/org/p2p/wallet/receive/eth/ReceiveEthereumFragment.kt
@@ -62,7 +62,6 @@ class EthereumReceiveFragment :
                 showUiKitSnackBar(messageResId = R.string.receive_eth_address_copied)
             }
             financeBlockAddress.apply {
-                updateHorizontalPadding(start = 16, end = 8)
                 bind(
                     FinanceBlockCellModel(
                         leftSideCellModel = LeftSideCellModel.IconWithText(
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/p2p/wallet/receive/eth/ReceiveEthereumFragment.kt
```diff
@@ -62,6 +62,7 @@ class EthereumReceiveFragment :
                 showUiKitSnackBar(messageResId = R.string.receive_eth_address_copied)
             }
             financeBlockAddress.apply {
+                updateHorizontalPadding(start = 16, end = 8)
```
kabramovp2p, line 65: what the... it should not be like that

it should be a common design

Davran331122: Screenshot 2023-03-12 at 23 03 42

Davran331122: we have 24 px space on right side, that's why I add this method x D

NikolaiEmelianov: add `app:layout_goneMarginEnd="0dp"` to `@+id/leftSideView` in widget_finance_block.xml

ui-kit/src/main/kotlin/org/p2p/uikit/components/finance_block/UiKitFinanceBlockView.kt
```diff
@@ -44,6 +44,13 @@ class UiKitFinanceBlockView @JvmOverloads constructor(
         bindViewStyle(styleType)
     }
 
+    fun updateHorizontalPadding(start: Int, end: Int) {
```
NikolaiEmelianov, line 47: not need if add layout_goneMarginEnd

NikolaiEmelianov (changes requested):

gslevinkov (approved):

NikolaiEmelianov (approved):

PR #1434, opened 2023-03-15
PWN-7380 - receive fixes

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7380

## Description of Work

PWN-7380 - receive fixes according to comments

### DIFF
--- a/app/src/main/java/org/p2p/wallet/receive/tokenselect/ReceiveTokensPresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/receive/tokenselect/ReceiveTokensPresenter.kt
@@ -1,5 +1,6 @@
 package org.p2p.wallet.receive.tokenselect
 
+import androidx.annotation.StringRes
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/p2p/wallet/receive/tokenselect/ReceiveTokensPresenter.kt
```diff
@@ -51,33 +51,33 @@ class ReceiveTokensPresenter(
...
-                    symbol = erc20Token.replaceTokenSymbol ?: token.tokenSymbol,
-                    iconUrl = erc20Token.tokenIconUrl ?: token.iconUrl,
-                    decimals = token.decimals,
-                    coingeckoId = erc20Token.coingeckoId,
-                    isWrapped = false,
-                    serumV3Usdc = null,
-                    serumV3Usdt = null
-                )
+        return ERC20Tokens.values().toMutableList()
+            .apply { removeAll(listOf(ERC20Tokens.LUNA, ERC20Tokens.UST)) }
```
gslevinkov, line 65: maybe filter will work better here?

ethereumkit/src/main/java/org/p2p/ethereumkit/external/model/ERC20Tokens.kt
```diff
@@ -50,43 +34,67 @@ enum class ERC20Tokens(
...
-        contractAddress = "0xbd31ea8212119f94a611fa969881cba3ea06fa3d",
-        mintAddress = "F6v4wfAdJB8D8p77bMXZgYt8TDKsYxLYxH5AFhUkYx9W",
-        coingeckoId = "terra-luna",
-        tokenIconUrl = null // TODO!
+    AVAX(
+        contractAddress = "0x85f138bfee4ef8e540890cfb48f620571d67eda3",
+        mintAddress = "KgV1GvrHQmRBY8sHQQeUKwTm2r2h8t4C8qt12Cw1HVE",
+        coingeckoId = "avalanche-2",
+        tokenIconUrl = "https://assets.coingecko.com/coins/images/12559/large/Avalanche_Circle_RedWhite_Trans.png?1670992574",
+        replaceTokenName = "Avalanche",
```
gslevinkov, line 61: no token symbol?

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/receive/tokenselect/ReceiveTokensPresenter.kt
kabramovp2p: can do with filter

just thought that removing it would be better than checking every item with filter

ethereumkit/src/main/java/org/p2p/ethereumkit/external/model/ERC20Tokens.kt
kabramovp2p: it means that we would not replace the real one

Davran331122 (approved):

PR #1455, opened 2023-03-20
PWN-7587 - THS update token each time login

NikolaiEmelianov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7587

## Description of Work

update token each time login

### DIFF
--- a/app/src/main/java/org/p2p/wallet/push_notifications/AppFirebaseMessagingService.kt
+++ b/app/src/main/java/org/p2p/wallet/push_notifications/AppFirebaseMessagingService.kt
@@ -11,6 +11,9 @@ import org.p2p.wallet.notification.AppNotificationManager
 import org.p2p.wallet.notification.FcmPushNotificationData
 import org.p2p.wallet.notification.NotificationType
 import timber.log.Timber
+import kotlinx.coroutines.launch
+import org.p2p.wallet.common.di.AppScope
+import org.p2p.wallet.push_notifications.ineractor.PushNotificationsInteractor
 
 private const val TAG = "AppFirebaseMessagingService"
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/p2p/wallet/push_notifications/AppFirebaseMessagingService.kt
```diff
@@ -19,10 +22,13 @@ class AppFirebaseMessagingService : FirebaseMessagingService(), KoinComponent {
     private val appNotificationManager: AppNotificationManager by inject()
     private val appsFlyerService: AppsFlyerService by inject()
     private val intercomPushService: IntercomPushService by inject()
+    private val pushNotificationsInteractor: PushNotificationsInteractor by inject()
+    private val appScope: AppScope by inject()
 
     override fun onNewToken(token: String) {
         appsFlyerService.onNewToken(token)
         intercomPushService.registerForPush(token)
+        appScope.launch { pushNotificationsInteractor.updateDeviceToken() }
```
gslevinkov, line 31: does onNewToken is called on every app launch?

NikolaiEmelianov: no

NikolaiEmelianov: signInByPinCode, signInByBiometric does

gslevinkov: oh, thanks!

gslevinkov (approved):

PR #1459, opened 2023-03-21
PWN-7613 - show usdCet if user has no tokens

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7613

## Description of Work

PWN-7613 - show usdCet if user has no tokens

### DIFF
--- a/app/src/main/java/org/p2p/wallet/bridge/send/ui/BridgeSendPresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/bridge/send/ui/BridgeSendPresenter.kt
@@ -17,6 +17,7 @@ import org.p2p.core.utils.scaleShort
 import org.p2p.ethereumkit.external.model.ERC20Tokens
 import org.p2p.wallet.BuildConfig
 import org.p2p.wallet.R
+import org.p2p.wallet.bridge.send.mapper.SendUiMapper
 import org.p2p.wallet.common.di.AppScope
 import org.p2p.wallet.common.mvp.BasePresenter
 import org.p2p.wallet.feerelayer.model.FeePayerSelectionStrategy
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/p2p/wallet/bridge/send/ui/BridgeSendPresenter.kt
```diff
@@ -60,6 +61,9 @@ class BridgeSendPresenter(
     sendModeProvider: SendModeProvider
 ) : BasePresenter<BridgeSendContract.View>(), BridgeSendContract.Presenter {
 
+    private val sendUiMapper = SendUiMapper()
```
gslevinkov, line 64: inject?

app/src/main/java/org/p2p/wallet/bridge/send/ui/BridgeSendPresenter.kt
```diff
@@ -132,11 +136,11 @@ class BridgeSendPresenter(
     private fun setupInitialToken(view: BridgeSendContract.View) {
         launch {
             // We should find SOL anyway because SOL is needed for Selection Mechanism
-            val userTokens = userInteractor.getNonZeroUserTokens()
+            var userTokens = userInteractor.getNonZeroUserTokens().filter { it.mintAddress in supportedTokensMints }
```
gslevinkov, line 139: instead of var you can use .ifEmpty

kabramovp2p: hmmm good point will change it!

Davran331122 (approved):

PR #1460, opened 2023-03-21
PWN-7765 - Pending transactions for token history

Davran331122:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7765


app/src/main/java/org/p2p/wallet/history/interactor/HistoryInteractor.kt
```diff
@@ -12,12 +12,12 @@ class HistoryInteractor(
...
 ) {
-    suspend fun loadHistory(limit: Int, mintAddress: String?): HistoryPagingResult =
+    suspend fun loadHistory(limit: Int, mintAddress: String): HistoryPagingResult =
         withContext(coroutineDispatchers.io) {
             return@withContext historyServiceRepository.loadHistory(limit, mintAddress)
         }
 
-    suspend fun loadNextPage(limit: Int, mintAddress: String?): HistoryPagingResult =
+    suspend fun loadNextPage(limit: Int, mintAddress: String): HistoryPagingResult =
         withContext(coroutineDispatchers.io) {
```
gslevinkov, line 21: i think we have withContext in historyServiceRepository

app/src/main/java/org/p2p/wallet/history/model/HistoryPendingTransaction.kt
```diff
@@ -0,0 +1,6 @@
+package org.p2p.wallet.history.model
+
+class HistoryPendingTransaction(
+    val txSignature: String,
```
gslevinkov, line 4: txSignature is base58?

gslevinkov (approved):

NikolaiEmelianov (approved):

PR #1479, opened 2023-03-22
PWN-7606 - fixes for coingecko

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7606

## Description of Work

PWN-7606 - fixes for coingecko use single for TokenPricesRemoteRepository and use common PriceCacheRepository.kt to use same logic under the hood

### DIFF
--- a/core/src/main/java/org/p2p/core/pricecache/PriceCacheRepository.kt
+++ b/core/src/main/java/org/p2p/core/pricecache/PriceCacheRepository.kt
@@ -1,0 +1,23 @@
+package org.p2p.core.pricecache
+
+import java.math.BigDecimal
+import org.p2p.core.model.TokenPriceWithMark
+
+abstract class PriceCacheRepository {
+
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/p2p/core/pricecache/PriceCacheRepository.kt
```diff
@@ -0,0 +1,23 @@
+package org.p2p.core.pricecache
+
+import java.math.BigDecimal
+import org.p2p.core.model.TokenPriceWithMark
+
+abstract class PriceCacheRepository {
```
gslevinkov, line 6: why not composition?
you can create TokenPricesInMemoryRepository and just inject that
inheritance is evil >:(

gslevinkov (approved):

1sakov:
Let's include this ticket in relearse @kabramovp2p

Davran331122 (approved):

PR #1481, opened 2023-03-23
NO-TICKET-Fix develop branch

Davran331122:


app/src/main/java/org/p2p/wallet/user/repository/UserInMemoryRepository.kt
```diff
@@ -131,7 +129,7 @@ class UserInMemoryRepository(
         val tokenData: TokenData? = findTokenData(mintAddress)
         return if (tokenData != null) {
             val price = getPriceByTokenId(tokenData.coingeckoId)
-            tokenConverter.fromNetwork(tokenData, price)
+            TokenConverter.fromNetwork(tokenData, price)
```
gslevinkov, line 132: any problems with the field?

gslevinkov (approved):

NikolaiEmelianov (approved):

app/src/main/java/org/p2p/wallet/user/repository/UserInMemoryRepository.kt
Davran331122: Yeah you inject this object, but forgot to provide it to module )

Davran331122: it's was injected only in this class

PR #1487, opened 2023-03-23
PWN-7614 - bridge send signin transaction

NikolaiEmelianov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7614

## Description of Work

bridge send signin transaction


app/src/main/java/org/p2p/wallet/bridge/send/BridgeSendInteractor.kt
```diff
@@ -1,10 +1,59 @@
...
+            mint = mint,
+            amount = amount.toPlainString()
+        )
+        val userAccount = Account(tokenKeyProvider.keyPair)
+
+        val signedTransaction = relaySdkFacade.signTransaction(
+            transaction = sendTransaction.transaction.toBase58Instance(),
+            keyPair = userAccount.getEncodedKeyPair().toBase58Instance(),
+            // todo need it?
+            recentBlockhash = null
```
gslevinkov, line 50: if there is recentBlockhash already in sendTransaction then you don't need it 
if there is not - fetch it here and pass as a parameter

Davran331122 (approved):

app/src/main/java/org/p2p/wallet/bridge/send/BridgeSendInteractor.kt
NikolaiEmelianov: recentBlockhash not need, deleted todo

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/bridge/send/BridgeSendInteractor.kt
```diff
@@ -1,10 +1,105 @@
...
+        source: SolAddress,
+        recipient: EthAddress,
+        token: Token.Active,
+        amount: BigDecimal
+    ): String = withContext(dispatchers.io) {
+        val tokenMint = token.mintAddress
+        val userPublicKey = tokenKeyProvider.publicKey
+        val userWallet = SolAddress(userPublicKey)
+
+        val relayAccount = feeRelayerAccountInteractor.getUserRelayAccount()
```
1sakov, line 50: feeRelayerAccountInteractor.getRelayInfo().feePayerAddress

1sakov:
this will need to be changed when fee relayer will be updated

1sakov (changes requested):
this will need to be changed when fee relayer will be updated

1sakov (approved):

PR #1511, opened 2023-03-29
PWN-7891 PWN-7894 - SVL. Fix account generation

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7891

## Description of Work

- fix link generation

### DIFF
--- a/app/src/main/java/org/p2p/wallet/newsend/model/TemporaryAccount.kt
+++ b/app/src/main/java/org/p2p/wallet/newsend/model/TemporaryAccount.kt
@@ -6,6 +6,8 @@ import kotlinx.parcelize.Parcelize
 import org.p2p.solanaj.core.PublicKey
 import org.p2p.wallet.utils.toPublicKey
 
+private const val SEND_LINK_FORMAT = "t.key.app/"
+
 @Parcelize
 data class TemporaryAccount(
     val symbols: String,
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/p2p/wallet/svl/model/SendLinkGenerator.kt
```diff
@@ -37,8 +43,13 @@ object SendLinkGenerator {
     private fun generateSymbols(): List<String> {
         val regex = Regex(REGEX_LINK_ALLOWED_SYMBOLS)
         val random = Random.Default
-        return (1..SYMBOLS_COUNT)
-            .map { random.nextInt(ASCII_CHARACTERS_COUNT).toString() }
```
gslevinkov, line 41: random.nextInt(ASCII_CHARACTERS_COUNT).toString() returned Int as a String. To convert ASCII Int to Char we user Char

app/src/main/java/org/p2p/wallet/newsend/model/TemporaryAccount.kt
```diff
@@ -6,21 +6,17 @@ import kotlinx.parcelize.Parcelize
 import org.p2p.solanaj.core.PublicKey
 import org.p2p.wallet.utils.toPublicKey
 
+private const val SEND_LINK_FORMAT = "t.key.app/"
```
kabramovp2p, line 9: I would prefer to add this in Gradle strings or in xml strings and use same way to use one resource

for case if we will rename it for example

kabramovp2p (approved):

NikolaiEmelianov (approved):

PR #1515, opened 2023-03-29
Release/v2.5.1 to develop

gslevinkov:


NikolaiEmelianov (approved):

app/src/main/java/org/p2p/wallet/jupiter/ui/settings/presenter/JupiterSwapFeeBuilder.kt
```diff
@@ -16,64 +17,75 @@ import org.p2p.wallet.jupiter.repository.model.findTokenByMint
...
+            SwapSettingFeeBox(
+                amountLamports = amountLamports,
+                amountUsd = amountUsd,
+                token = lpToken
+            )
         }
     }
+
+    private suspend fun loadRateForToken(token: JupiterSwapToken): SwapRateLoaderState.Loaded? {
+        return swapStateManager.getTokenRate(SwapTokenModel.JupiterToken(token))
```
1sakov, line 80: Make sure that Swap will not be blocked if token rate will return exception

PR #1516, opened 2023-03-29
PWN-7635 - push analytics

NikolaiEmelianov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7635

## Description of Work

push analytics

PR #1520, opened 2023-03-30
NO-TICKET - Fix develop

gslevinkov:


NikolaiEmelianov:
fixes here https://github.com/p2p-org/key-app-android/pull/1516

PR #1528, opened 2023-03-31
PWN-7940 - SVL. Remove date subtitle in bottom sheet

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7940


app/src/main/java/org/p2p/wallet/svl/interactor/ReceiveViaLinkInteractor.kt
```diff
@@ -22,8 +23,15 @@ class ReceiveViaLinkInteractor(
     private val tokenKeyProvider: TokenKeyProvider
 ) {
 
+    private class ReceiveViaLinkError(
+        override val message: String,
+        override val cause: Throwable? = null
+    ) : Throwable()
```
1sakov, line 29: move to a separate file

gslevinkov: Not used anywhere than here, are you sure?

kabramovp2p (approved):

1sakov (approved):

app/src/main/java/org/p2p/wallet/svl/interactor/ReceiveViaLinkInteractor.kt
gslevinkov: Not at all, it's for Sentry and Firebase for specific details in issue list

PR #1530, opened 2023-03-31
PWN-7901 - Add withTld to name-service

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PW-7901

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -171,11 +171,10 @@ dependencies {
     testImplementation("io.insert-koin:koin-test:$koinVersion")
     testImplementation("io.insert-koin:koin-test-junit4:$koinVersion")
 
-    testImplementation("org.assertj:assertj-core:3.22.0")
+    testImplementation("com.willowtreeapps.assertk:assertk-jvm:0.25")
     testImplementation("org.slf4j:slf4j-nop:1.7.30")
 
     testImplementation("org.junit.jupiter:junit-jupiter-api:5.8.2")
-    testImplementation("org.assertj:assertj-core:3.22.0")
 
     testRuntimeOnly("org.junit.platform:junit-platform-launcher:1.8.2") {
         because("This lib comes shipped with the IDE and it possible that newer versions of JUnit 5 maybe be incompatible with the version of junit-platform-launcher shipped with the IDE.")
--- a/app/src/main/java/org/p2p/wallet/auth/model/Username.kt
+++ b/app/src/main/java/org/p2p/wallet/auth/model/Username.kt
@@ -3,6 +3,5 @@ package org.p2p.wallet.auth.model
 data class Username(
     val value: String,
     val domainPrefix: String,
-    // FIXME: New version of name service returns domain as well. We should remove domainPrefix asap
     val fullUsername: String = value + domainPrefix
 )
--- a/app/src/main/java/org/p2p/wallet/auth/model/Username.kt
+++ b/app/src/main/java/org/p2p/wallet/auth/model/Username.kt
@@ -1,7 +1,11 @@
 package org.p2p.wallet.auth.model
 
+/**
+ * @see [org.p2p.wallet.auth.username.repository.UsernameParser]
+ */
 data class Username(
     val value: String,
     val domainPrefix: String,
+) {
     val fullUsername: String = value + domainPrefix
-)
+}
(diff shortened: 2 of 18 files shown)


kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/auth/model/Username.kt
```diff
@@ -3,6 +3,5 @@ package org.p2p.wallet.auth.model
 data class Username(
     val value: String,
     val domainPrefix: String,
-    // FIXME: New version of name service returns domain as well. We should remove domainPrefix asap
     val fullUsername: String = value + domainPrefix
```
1sakov, line 6: we better remove this and `domainPrefix` fields. 
All usernames will come from backend. 
And looks like we need to remove feature toggle for `username_value`
Please workaround

app/build.gradle.kts
```diff
@@ -171,11 +171,10 @@ dependencies {
     testImplementation("io.insert-koin:koin-test:$koinVersion")
     testImplementation("io.insert-koin:koin-test-junit4:$koinVersion")
 
-    testImplementation("org.assertj:assertj-core:3.22.0")
+    testImplementation("com.willowtreeapps.assertk:assertk-jvm:0.25")
```
gslevinkov, line 174: AssertJ erases generics when using assertThat, making assertions very bad. 
assertK has the same API

app/src/main/java/org/p2p/wallet/auth/model/Username.kt
gslevinkov: done!
checked on main screen, search fragment

app/src/test/java/org/p2p/wallet/auth/username/repository/UsernameParserTest.kt
```diff
@@ -0,0 +1,85 @@
...
+import io.mockk.every
+import io.mockk.mockk
+import org.junit.Test
+import org.p2p.core.utils.Constants
+import org.p2p.wallet.auth.model.Username
+import org.p2p.wallet.auth.username.repository.model.UsernameDetails
+import org.p2p.wallet.common.feature_toggles.toggles.remote.UsernameDomainFeatureToggle
+import org.p2p.wallet.utils.toBase58Instance
+
+class UsernameParserTest {
```
gslevinkov, line 16: unit test for username parsing

app/src/main/java/org/p2p/wallet/auth/model/Username.kt
1sakov: I can't see, if it's done. 
You are still concatenating here: `value + domainPrefix`

app/src/main/java/org/p2p/wallet/auth/username/repository/UsernameParser.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.p2p.wallet.utils.Base58String
+
+/**
+ * @see [org.p2p.wallet.auth.username.repository.UsernameParserTest]
+ */
+class UsernameParser(
+    private val usernameDomainFeatureToggle: UsernameDomainFeatureToggle
+) {
+    /**
+     * @param anyUsername - username with domain prefix or not
```
1sakov, line 15: Did you ask if we really need this logic at all? 
I don't think feature toggle is needed, moreover the domain like `.key.sol`. 

Our business couldn't own this domain as planned, so looks like the feature toggle and this logic is redundant at all

app/src/test/java/org/p2p/wallet/auth/username/repository/UsernameParserTest.kt
```diff
@@ -0,0 +1,85 @@
...
+                prop(UsernameDetails::ownerAddress).isEqualTo(ownerAddress)
+
+                val username = prop(UsernameDetails::username)
+                username.prop(Username::value).isEqualTo(expectedUsername)
+                username.prop(Username::domainPrefix).isEqualTo(expectedDomain)
+                username.prop(Username::fullUsername).isEqualTo(expectedUsername + expectedDomain)
+            }
+        }
+    }
+}
```
1sakov, line 85: looks nice!

1sakov (approved):
minor comments

PR #1563, opened 2023-04-11
PWN-7979 - don't refresh the tx history when the link history is closed

eduardmaximovich:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7979

## Description of Work

Adding `links history` fragment instead of replacing it.

## Screenshots (Optional)

### DIFF
--- a/app/src/main/java/org/p2p/wallet/history/ui/history/HistoryFragment.kt
+++ b/app/src/main/java/org/p2p/wallet/history/ui/history/HistoryFragment.kt
@@ -13,6 +13,7 @@ import org.p2p.wallet.history.ui.historylist.HistoryListViewType
 import org.p2p.wallet.history.ui.sendvialink.HistorySendLinksFragment
 import org.p2p.wallet.moonpay.ui.new.NewBuyFragment
 import org.p2p.wallet.moonpay.ui.transaction.SellTransactionDetailsBottomSheet
+import org.p2p.wallet.utils.addFragment
 import org.p2p.wallet.utils.replaceFragment
 import org.p2p.wallet.utils.viewbinding.viewBinding
 
@@ -38,7 +39,7 @@ class HistoryFragment :
     }
 
     override fun onUserSendLinksClicked() {
-        replaceFragment(HistorySendLinksFragment.create())
+        addFragment(HistorySendLinksFragment.create())
     }
 
     override fun onTransactionClicked(transactionId: String) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/p2p/wallet/history/ui/history/HistoryFragment.kt
```diff
@@ -38,7 +39,7 @@ class HistoryFragment :
     }
 
     override fun onUserSendLinksClicked() {
-        replaceFragment(HistorySendLinksFragment.create())
+        addFragment(HistorySendLinksFragment.create())
```
kabramovp2p, line 42: why we need to use add here?

eduardmaximovich: to prevent destroying the HistoryFragment, we don't need to reload the transaction history from scratch as a result

1sakov: I propose to add check in `presenter` if history is not null or empty, then we don't load the history. 
And we will still be able to leave this as `replace`

eduardmaximovich: @1sakov The HistoryListPresenter is destroyed when we replace the HistoryFragment with the links fragment because it attaches directly to the HistoryListView. As the HistoryListView is also being destroyed, we can't simply check whether the data has already been loaded or not within the presenter. Or am I missing something?

1sakov (approved):

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryFragment.kt
```diff
@@ -26,11 +27,13 @@ class HistoryFragment :
     }
 
     override val presenter: HistoryContract.Presenter by inject()
+    private val historyListPresenter: HistoryListViewContract.Presenter by inject()
```
gslevinkov, line 30: could you leave a comment on why we need to inject this presenter outside of view?

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/history/ui/history/HistoryFragment.kt
eduardmaximovich: Decided to move the HistoryListPresenter back to the HistoryFragment as it was before, cause the presenter should survive destroying HistoryListView

PR #1579, opened 2023-04-17
PWN-8112 - Fix Home Screen buttons and history screen loading

Davran331122:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8112
https://p2pvalidator.atlassian.net/browse/PWN-8150


app/src/main/java/org/p2p/wallet/home/ui/main/HomePresenter.kt
```diff
@@ -135,6 +139,14 @@ class HomePresenter(
         }
     }
 
+    private fun attachToActionButtons() {
```
gslevinkov, line 142: observeActionButtonsState

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/home/ui/main/HomePresenter.kt
eduardmaximovich: +

eduardmaximovich (approved):

PR #1591, opened 2023-04-18
PWN-7811 - added support for choosing derived account

eduardmaximovich:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7811

## Description of Work

- Added support for choosing derived account
- Added ripple to account foreground
- Default wallet index = 0 for deprecated code

### DIFF
--- a/app/src/main/java/org/p2p/wallet/restore/ui/derivable/DerivableAccountsAdapter.kt
+++ b/app/src/main/java/org/p2p/wallet/restore/ui/derivable/DerivableAccountsAdapter.kt
@@ -14,7 +14,11 @@ import org.p2p.wallet.restore.model.DerivableAccount
 private const val FULL_ALPHA = 1.0f
 private const val HALF_ALPHA = 0.5f
 
-class DerivableAccountsAdapter : RecyclerView.Adapter<RecyclerView.ViewHolder>() {
+typealias OnAccountClick = (walletIndex: Int) -> Unit
+
+class DerivableAccountsAdapter(
+    private val onAccountClick: OnAccountClick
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/p2p/wallet/restore/ui/derivable/DerivableAccountsAdapter.kt
```diff
@@ -27,22 +31,22 @@ class DerivableAccountsAdapter : RecyclerView.Adapter<RecyclerView.ViewHolder>()
     override fun getItemCount(): Int = data.size
 
     override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): RecyclerView.ViewHolder =
-        ViewHolder(parent)
+        ViewHolder(
+            binding = ItemTokenSimpleBinding.inflate(
```
gslevinkov, line 35: binding = parent.inflateViewBinding(attachToParent = false)

gslevinkov: but we extract creating binding inside the constructor, like org.p2p.wallet.home.ui.new.adapter.CategoryTitleViewHolder

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/restore/ui/derivable/DerivableAccountsAdapter.kt
eduardmaximovich: yeah, that's a great example, it's much better than having separate constructor to bind view as it was before

PR #1595, opened 2023-04-19
PWN-7948 - SVL. Additional history fixes

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-7948

### DIFF
--- a/app/src/main/java/org/p2p/wallet/svl/model/SendViaLinkClaimingState.kt
+++ b/app/src/main/java/org/p2p/wallet/svl/model/SendViaLinkClaimingState.kt
@@ -27,10 +27,18 @@ sealed interface SendViaLinkClaimingState {
         @DrawableRes val iconRes: Int
     ) : SendViaLinkClaimingState {
 
-        constructor() : this(
-            titleRes = R.string.send_via_link_error_failed_parsing_title,
-            subTitleRes = null,
-            iconRes = R.drawable.ic_not_found
-        )
+        companion object {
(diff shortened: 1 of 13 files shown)


kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/svl/model/SendViaLinkClaimingState.kt
```diff
@@ -27,10 +27,18 @@ sealed interface SendViaLinkClaimingState {
...
-            iconRes = R.drawable.ic_not_found
-        )
+        companion object {
+            val UNKNOWN_ERROR = ParsingFailed(
+                titleRes = R.string.send_via_link_error_failed_parsing_title,
+                subTitleRes = null,
+                iconRes = R.drawable.ic_not_found
+            )
+
+            val INTERNET_ERROR = ParsingFailed(
```
1sakov, line 37: That looks weird. 
Mb better make them as functions? 

Smth like `buildNetworkError(): ParsingFailed`, `buildUnknownError(): ParsingFailed`
What do you think?

1sakov (approved):

app/src/main/java/org/p2p/wallet/svl/model/SendViaLinkClaimingState.kt
gslevinkov: Will do, thanks!

PR #1614, opened 2023-04-21
PWN-8178 - remove sol from supported tokens

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8178

## Description of Work

PWN-8178 - remove sol from supported tokens


ethereumkit/src/main/java/org/p2p/ethereumkit/external/model/ERC20Tokens.kt
```diff
@@ -100,6 +91,7 @@ enum class ERC20Tokens(
     companion object {
 
         const val ETH_DECIMALS = 18
+        const val SOL_TOKEN_URL = "https://assets.coingecko.com/coins/images/4128/large/solana.png?1640133422"
```
gslevinkov, line 94: I don't think it's connected with ERC20 tokens. Maybe Const class is better?

gslevinkov (approved):

ethereumkit/src/main/java/org/p2p/ethereumkit/external/model/ERC20Tokens.kt
kabramovp2p: it is used only for bridges
but we can move it to Constants

but anyway it is partially connected =)

PR #1630, opened 2023-04-25
PWN-8222 - Fix token sorting on main screen before PTR

Davran331122:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8222


gslevinkov (approved):

eduardmaximovich (approved):

app/src/main/java/org/p2p/wallet/user/interactor/UserInteractor.kt
```diff
@@ -154,7 +154,7 @@ class UserInteractor(
             val price = prices.find { it.tokenId == token.coingeckoId }
             val totalInUsd = price?.price?.let { token.total.times(it) }
             token.copy(rate = price?.price, totalInUsd = totalInUsd)
-        }
+        }.sortedWith(TokenComparator())
```
1sakov, line 157: We are sorting before saving into DB. 
Are you sure the `map` here breaks the sorting?

1sakov (changes requested):

PR #1639, opened 2023-04-27
PWN-4179 - remove wSol (wrapped sol) from tokens list

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-4179

## Description of Work

PWN-4179 - remove wSol (wrapped sol) from tokens list

### DIFF
--- a/app/src/main/java/org/p2p/wallet/receive/list/TokenListPresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/receive/list/TokenListPresenter.kt
@@ -1,6 +1,7 @@
 package org.p2p.wallet.receive.list
 
 import kotlinx.coroutines.launch
+import org.p2p.core.utils.Constants
 import org.p2p.wallet.common.analytics.interactor.ScreensAnalyticsInteractor
 import org.p2p.wallet.common.mvp.BasePresenter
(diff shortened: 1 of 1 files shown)


Davran331122 (approved):

app/src/main/java/org/p2p/wallet/receive/list/TokenListPresenter.kt
```diff
@@ -50,7 +51,8 @@ class TokenListPresenter(
         launch {
             interactor.getTokenListFlow().collect { data ->
                 if (data.result.isNotEmpty()) {
-                    view?.showItems(data.result, scrollToUp)
+                    val filteredResult = data.result.filter { it.mintAddress != Constants.WRAPPED_SOL_MINT }
```
gslevinkov, line 54: You can put filter on flow, maybe it will optimise things a bit

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/receive/list/TokenListPresenter.kt
kabramovp2p: this flow is reused in new tokens list

PR #1644, opened 2023-04-28
PWN-8212 - Legacy FT. 2. Old settings screen

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8212


app/src/main/java/org/p2p/wallet/settings/SettingsModule.kt
```diff
@@ -1,67 +1,35 @@
...
 
 object SettingsModule : InjectionModule {
 
     override fun create() = module {
         factoryOf(::SettingsInteractor)
         factoryOf(::ThemeInteractor)
+
         factoryOf(::SettingsItemMapper)
-        factory {
-            // holy shit, TODO smth with this dependencies!
```
gslevinkov, line 37: did something with with these dependencies :))

Davran331122 (approved):

Davran331122 (approved):

app/src/main/java/org/p2p/wallet/settings/SettingsModule.kt
eduardmaximovich: that was easy :D

kabramovp2p (approved):

eduardmaximovich (approved):

PR #1659, opened 2023-05-03
PWN-8271 - Add progress dialog for claiming bundles

Davran331122:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8271

### DIFF
--- a/app/src/main/java/org/p2p/wallet/bridge/claim/ClaimModule.kt
+++ b/app/src/main/java/org/p2p/wallet/bridge/claim/ClaimModule.kt
@@ -6,6 +6,7 @@ import org.koin.dsl.bind
 import org.koin.dsl.module
 import org.p2p.wallet.bridge.claim.interactor.ClaimInteractor
 import org.p2p.wallet.bridge.claim.mapper.EthereumBundleMapper
+import org.p2p.wallet.bridge.claim.repository.EthereumClaimLocalRepository
 import org.p2p.wallet.bridge.claim.repository.EthereumClaimRemoteRepository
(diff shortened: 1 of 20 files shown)


kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/bridge/claim/ClaimModule.kt
```diff
@@ -17,6 +18,7 @@ object ClaimModule : InjectionModule {
 
     override fun create() = module {
         singleOf(::EthereumClaimRemoteRepository) bind EthereumClaimRepository::class
+        singleOf(::EthereumClaimLocalRepository)
```
gslevinkov, line 21: where is the interface?

app/src/main/java/org/p2p/wallet/home/ui/main/HomePresenter.kt
```diff
@@ -109,7 +112,9 @@ class HomePresenter(
 
     init {
         val userSeedPhrase = seedPhraseProvider.getUserSeedPhrase().seedPhrase
-        ethereumInteractor.setup(userSeedPhrase = userSeedPhrase)
+        if (userSeedPhrase.isNotEmpty()) {
```
gslevinkov, line 115: please put else branch here and log some Throwable for us to know how many users don't have user seed phrase like this: 
if (userSeedPhrase.isNotEmpty(){
...
} else {
Timber.e(NoSeedPhraseForEth(), "ETH is not init, no seed phrase")
}

gslevinkov (approved):

PR #1683, opened 2023-05-08
PWN-8245 - show rates loading warning in snackbar how it was before

eduardmaximovich:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8245

## Description of Work

Reverted the snackbar error that occurs when unable to load token rates.
Originally, it was accidentially removed down there [link](https://github.com/p2p-org/key-app-android/pull/1547/files#diff-7d20ada08a280cc7d4172bc771d6c4a73602cb105ebd010f0f0aba4425fcc229L404)

### DIFF
--- a/app/src/main/java/org/p2p/wallet/home/ui/main/HomePresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/home/ui/main/HomePresenter.kt
@@ -132,7 +132,14 @@ class HomePresenter(
                 view?.showRefreshing(true)
                 userInteractor.loadAllTokensDataIfEmpty()
                 val solTokens = userInteractor.loadUserTokensAndUpdateLocal(tokenKeyProvider.publicKey.toPublicKey())
-                async { userInteractor.loadUserRates(solTokens) }
+                async {
+                    try {
+                        userInteractor.loadUserRates(solTokens)
+                    } catch (t: Throwable) {
+                        Timber.e(t, "Error on loading Sol rates")
+                        view?.showUiKitSnackBar(messageResId = R.string.error_token_rates)
+                    }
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/p2p/wallet/home/ui/main/HomePresenter.kt
```diff
@@ -132,7 +132,14 @@ class HomePresenter(
                 view?.showRefreshing(true)
                 userInteractor.loadAllTokensDataIfEmpty()
                 val solTokens = userInteractor.loadUserTokensAndUpdateLocal(tokenKeyProvider.publicKey.toPublicKey())
-                async { userInteractor.loadUserRates(solTokens) }
+                async {
+                    try {
+                        userInteractor.loadUserRates(solTokens)
+                    } catch (t: Throwable) {
+                        Timber.e(t, "Error on loading user rates")
```
gslevinkov, line 139: no need to Timber.e this stuff, only Timber.i

gslevinkov (approved):

gslevinkov (approved):

gslevinkov:
no task number in the description

gslevinkov:
let's put this fix in 2.7.1

app/src/main/java/org/p2p/wallet/home/ui/main/HomePresenter.kt
Davran331122: try catch will not handle error without using `await()`

Davran331122 (changes requested):

Davran331122 (approved):

PR #1690, opened 2023-05-10
PWN-8333 - Token history socket integration

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8333

### DIFF
--- a/app/src/main/java/org/p2p/wallet/history/ui/token/TokenHistoryFragment.kt
+++ b/app/src/main/java/org/p2p/wallet/history/ui/token/TokenHistoryFragment.kt
@@ -6,7 +6,6 @@ import android.os.Bundle
 import android.view.View
 import org.koin.android.ext.android.inject
 import org.koin.core.parameter.parametersOf
-import timber.log.Timber
 import org.p2p.core.token.Token
 import org.p2p.core.utils.Constants
 import org.p2p.wallet.BuildConfig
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/p2p/wallet/history/ui/token/TokenHistoryFragment.kt
```diff
@@ -115,6 +114,16 @@ class TokenHistoryFragment :
         }
     }
 
+    override fun renderTokenAmounts(token: Token.Active) {
+        binding.totalTextView.text = token.getFormattedTotal(includeSymbol = true)
+        binding.usdTotalTextView.text = token.getFormattedUsdTotal()
+    }
+
+    override fun loadTokenHistoryList() {
+        Timber.e("Fetching history")
```
Davran331122, line 123: Oops

Davran331122 (approved):

app/src/main/java/org/p2p/wallet/history/ui/token/TokenHistoryFragment.kt
kabramovp2p: yes remove this pls

gslevinkov: thanks guys!

PR #1699, opened 2023-05-11
PWN-8333 - extracted creation to factory, fixes in parameters, more abstractions

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8333

## Description of Work
- factory for creating interface instead of implementation. allows to switch impl faster in future if needed
- factory incapsulated creation logic making WebSocketClient more cleaner
- some renaming that I forgot to do
- params fixes, so updates come faster
- wrapping all errors in own class to distinct them in Sentry

### DIFF
--- a/solana/src/main/java/org/p2p/solanaj/ws/SocketClientCreateResult.kt
+++ b/solana/src/main/java/org/p2p/solanaj/ws/SocketClientCreateResult.kt
@@ -1,0 +1,6 @@
+package org.p2p.solanaj.ws
+
+sealed interface SocketClientCreateResult {
+    class Created(val instance: SubscriptionSocketClient) : SocketClientCreateResult
+    class Failed(override val cause: Throwable) : Throwable(), SocketClientCreateResult
+}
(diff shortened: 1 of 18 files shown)


solana/src/main/java/org/p2p/solanaj/ws/SocketClientCreateResult.kt
```diff
@@ -0,0 +1,6 @@
+package org.p2p.solanaj.ws
+
+sealed interface SocketClientCreateResult {
```
Davran331122, line 3: For me it's not obvious to have wrapped instance.
Why did you recognize to use it in this way ?

Davran331122 (approved):

solana/src/main/java/org/p2p/solanaj/ws/SocketClientCreateResult.kt
gslevinkov: That's a whole subject to discuss
1) Using a monad wrapper like functional `Either` or `Result` or `Optional` is a better way of doing things instead of implicitly throwing exceptions inside
2) There is a guarantee for you, as a user that this method won't throw anything unexpected and you can remove unnecessary try-catch inside your code
3) This result is explicit - you see that this method can return an instance, or error, and you are forced by the complier to handle this type of thing. sealed result type provides Types and Compile Time Safety

https://elizarov.medium.com/kotlin-and-exceptions-8062f589d07

PR #1710, opened 2023-05-12
PWN-8329 - fixed loading token rates when connection is bad

eduardmaximovich:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8329

## Description of Work

- Fixed loading token rates when connection error is present

### DIFF
--- a/app/src/main/java/org/p2p/wallet/home/ui/main/HomePresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/home/ui/main/HomePresenter.kt
@@ -138,14 +138,18 @@ class HomePresenter(
     }
 
     private fun loadSolanaTokens() {
+        launch {
+            // ignore internet state; tokens must display regardless of any error
+            attachToPollingTokens()
+        }
         launchInternetAware(connectionManager) {
             try {
                 view?.showRefreshing(true)
                 userInteractor.loadAllTokensDataIfEmpty()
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/p2p/wallet/home/ui/main/HomePresenter.kt
```diff
@@ -167,6 +170,23 @@ class HomePresenter(
         }
     }
 
+    override fun refreshTokens() {
```
eduardmaximovich, line 173: I've just moved it from bottom to top

app/src/main/java/org/p2p/wallet/user/interactor/UserInteractor.kt
```diff
@@ -122,6 +122,20 @@ class UserInteractor(
         }
     }
 
+    suspend fun loadUserRatesIfEmpty(userTokens: List<Token.Active>) {
+        when {
+            // if prices are not loaded at all, load them
+            !userLocalRepository.arePricesLoaded() -> {
+                loadUserRates(userTokens)
```
gslevinkov, line 129: do you want to try catch that? Error while loading user rates should not block other request and startPolling()

gslevinkov (approved):

1sakov (approved):

PR #1716, opened 2023-05-15
PWN-8493 - Validate ALL swap routes

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8493

## Description of Work

- instead of validating route by route, we are doing async jobs to validate them all. Results are astonishing fast!

### DIFF
--- a/solana/src/main/java/org/p2p/solanaj/model/types/Encoding.kt
+++ b/solana/src/main/java/org/p2p/solanaj/model/types/Encoding.kt
@@ -9,7 +9,7 @@ enum class Encoding(
     @SerializedName("base64")
     BASE64("base64"),
 
-    @Deprecated("Slow for solana, use Base 58")
+    @Deprecated("Slow for solana, use Base 64")
     @SerializedName("base58")
     BASE58("base58");
 }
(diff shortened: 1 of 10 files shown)


kabramovp2p (approved):

solana/src/main/java/org/p2p/solanaj/model/types/Encoding.kt
```diff
@@ -9,6 +9,7 @@ enum class Encoding(
     @SerializedName("base64")
     BASE64("base64"),
 
+    @Deprecated("Slow for solana, use Base 58")
```
eduardmaximovich, line 12: Maybe you meant `use base 64`? Or possibly this deprecation should be over `BASE64`

eduardmaximovich (approved):

solana/src/main/java/org/p2p/solanaj/model/types/Encoding.kt
gslevinkov, line 12:
```suggestion
    @Deprecated("Slow for solana, use Base 64")
```

gslevinkov: fixed, thanks!

PR #1721, opened 2023-05-16
PWN-8400 - filter transactions that are not in process to show

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8400

## Description of Work

PWN-8400 - filter transactions that are not in process to show

### DIFF
--- a/app/src/main/java/org/p2p/wallet/history/model/rpc/RpcHistoryTransaction.kt
+++ b/app/src/main/java/org/p2p/wallet/history/model/rpc/RpcHistoryTransaction.kt
@@ -334,6 +334,7 @@ sealed class RpcHistoryTransaction(
         val tokenSymbol: String,
         val iconUrl: String?,
         val fees: List<RpcFee>?,
+        val claimKey: String
     ) : RpcHistoryTransaction(date, signature, blockNumber, status, type) {
 
         @StringRes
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/p2p/wallet/history/model/rpc/RpcHistoryTransaction.kt
```diff
@@ -378,6 +379,7 @@ sealed class RpcHistoryTransaction(
         val tokenSymbol: String,
         val iconUrl: String?,
         val fees: List<RpcFee>?,
+        val message: String
```
gslevinkov, line 382: why not bridgeServiceKey?

app/src/main/java/org/p2p/wallet/sell/interactor/HistoryItemMapper.kt
```diff
@@ -76,6 +80,18 @@ class HistoryItemMapper(
...
 
+    private fun RpcHistoryTransaction.isNotProcessing(filterBundleIds: List<String>): Boolean {
+        return when (this) {
+            is RpcHistoryTransaction.WormholeReceive -> {
+                claimKey !in filterBundleIds
+            }
+            is RpcHistoryTransaction.WormholeSend -> {
+                message !in filterBundleIds
+            }
+            else -> true
```
gslevinkov, line 91:
```suggestion
            is RpcHistoryTransaction.WormholeReceive -> claimKey !in filterBundleIds
            is RpcHistoryTransaction.WormholeSend -> message !in filterBundleIds
            else -> true
```

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/history/model/rpc/RpcHistoryTransaction.kt
kabramovp2p: because in transaction it is `message` and in SendDetails it is `id`
third option will just make more confuse =(

PR #1722, opened 2023-05-16
PWN-8248 - Update labels when view created again

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8248

### DIFF
--- a/app/src/main/java/org/p2p/wallet/newsend/model/CalculationMode.kt
+++ b/app/src/main/java/org/p2p/wallet/newsend/model/CalculationMode.kt
@@ -19,6 +19,8 @@ import org.p2p.core.utils.toUsd
 import org.p2p.wallet.infrastructure.network.provider.SendModeProvider
 import org.p2p.wallet.utils.divideSafe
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/p2p/wallet/newsend/model/CalculationMode.kt
```diff
@@ -51,16 +53,17 @@ class CalculationMode(
     private var usdAmount: BigDecimal = BigDecimal.ZERO
 
     fun updateToken(newToken: Token.Active) {
-        if (::token.isInitialized && newToken.mintAddress == this.token.mintAddress) return
-        this.token = newToken
+        if (::token.isInitialized && newToken.mintAddress == this.token.mintAddress) {
```
gslevinkov, line 56: when coming back from other screens in Send flow, updateToken was called, but didn't call updateLabels()

kabramovp2p (approved):

eduardmaximovich (approved):

PR #1726, opened 2023-05-17
PWN-8248 - Update switch buttons when view is created again

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8248

### DIFF
--- a/app/src/main/java/org/p2p/wallet/newsend/ui/NewSendPresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/newsend/ui/NewSendPresenter.kt
@@ -145,6 +145,7 @@ class NewSendPresenter(
         launch {
             view.showToken(token)
             calculationMode.updateToken(token)
+            checkTokenRatesAndSetSwitchAmountState(token)
 
             val userTokens = userInteractor.getNonZeroUserTokens()
             val isTokenChangeEnabled = userTokens.size > 1 && selectedToken == null


app/src/main/java/org/p2p/wallet/newsend/ui/NewSendPresenter.kt
```diff
@@ -145,6 +145,7 @@ class NewSendPresenter(
         launch {
             view.showToken(token)
             calculationMode.updateToken(token)
+            checkTokenRatesAndSetSwitchAmountState(token)
```
gslevinkov, line 148: again, on restore view switch amount state was not init

1sakov (approved):

app/src/main/java/org/p2p/wallet/newsend/SendFeeRelayerManager.kt
```diff
@@ -145,10 +147,10 @@ class SendFeeRelayerManager(
                         feeRelayerFee = feeState.feeInSol
                     )
                     currentState = UpdateFee(solanaFee = solanaFee, feeLimitInfo = feeLimitInfo)
-                    sendInteractor.setFeePayerToken(solToken)
+                    sendInteractor.switchFeePayerToSol(solToken)
                 }
                 is FeeCalculationState.Success -> {
-                    sendInteractor.setFeePayerToken(sourceToken)
+                    sendInteractor.setFeePayerToken(feePayer)
```
gslevinkov, line 153: i've tried to pass feePayer token instead of sourceToken because it doesn't make sense to always pass source token

app/src/main/java/org/p2p/wallet/newsend/model/SendSolanaFee.kt
```diff
@@ -161,18 +163,39 @@ data class SendSolanaFee constructor(
...
+            appendLine("alternativeFeePayerTokens = ${alternativeFeePayerTokens.map(Token.Active::tokenSymbol)}")
+            appendLine("isValidToSwitchOnSource = $isValidToSwitchOnSource")
+            appendLine("shouldSwitchToSpl = $shouldSwitchToSpl")
+        })
         return when {
-            // if there is enough SPL token balance to cover amount and fee
-            !isSourceSol && sourceTokenTotal.isMoreThan(totalNeeded) &&
-                isValidToSwitchOnSource -> FeePayerState.SwitchToSpl(sourceToken)
-            hasAlternativeFeePayerTokens -> FeePayerState.SwitchToSpl(alternativeFeePayerTokens.first())
+            shouldSwitchToSpl -> {
```
gslevinkov, line 184: the main problem is here
User selected AVA token to send to account a.p2p.sol
a.p2p.sol doesn't have token account so user needs to pay fee to create one
we switch fee payer token to SOL
after returning to main send screen we see that fee is calculated in SOL
but when tapping on SWITCH_MODE button, fee_payer token is switched to USDC 

in this case I saw that 
isEnoughSolBalance = true
hasAlternativeFeePayerTokens = true
and that's why in when case SendSolanaFee selects new fee payer token = USDC

PR #1782, opened 2023-05-31
PWN-8452 - Striga API. resend_sms

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8452

## Description of Work

- added resend sms method
- added doc to API to understand what errors to catch

### DIFF
--- a/app/src/main/java/org/p2p/wallet/striga/model/StrigaDataLayerError.kt
+++ b/app/src/main/java/org/p2p/wallet/striga/model/StrigaDataLayerError.kt
@@ -10,8 +10,11 @@ sealed class StrigaDataLayerError(override val message: String) : Throwable() {
     ) : StrigaDataLayerError("Striga API unavailable, code: ${cause.code()} body: $body")
 
     class ApiServiceError(
-        val response: StrigaApiErrorResponse
-    ) : StrigaDataLayerError("Striga API returned error: code=${response.errorCode?.code} details=${response.details}")
+        val response: StrigaApiErrorResponse,
+    ) : StrigaDataLayerError("Striga API returned error: code=${response.errorCode.code} details=${response.details}") {
+        val errorCode: StrigaApiErrorCode
+            get() = response.errorCode
+    }
 
     class InternalError(
         override val cause: Throwable? = null,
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/p2p/wallet/striga/user/api/StrigaApi.kt
```diff
@@ -9,6 +9,19 @@ interface StrigaApi {
...
+     * 30044 - mobile already in use
+     * 30031 - invalid verification code
+     * 30003 - exceeded verification attempts
+     * 30005 - user doesn't exist, critical error
+     */
     @POST("v1/user/verify-mobile")
     suspend fun verifyMobileNumber(@Body body: StrigaVerifyMobileNumberRequest)
+
+    /**
+     * 31009 - mobile already in use
```
gslevinkov, line 22: we have StrigaApiErrorCode to handle this errors
I thought it would be better for you as a users of this api to know what to catch

app/src/main/java/org/p2p/wallet/striga/model/StrigaDataLayerError.kt
```diff
@@ -10,8 +10,11 @@ sealed class StrigaDataLayerError(override val message: String) : Throwable() {
     ) : StrigaDataLayerError("Striga API unavailable, code: ${cause.code()} body: $body")
 
     class ApiServiceError(
-        val response: StrigaApiErrorResponse
-    ) : StrigaDataLayerError("Striga API returned error: code=${response.errorCode?.code} details=${response.details}")
+        val response: StrigaApiErrorResponse,
+    ) : StrigaDataLayerError("Striga API returned error: code=${response.errorCode.code} details=${response.details}") {
+        val errorCode: StrigaApiErrorCode
+            get() = response.errorCode
```
1sakov, line 16: we don't need `get` here since the response cannot be changed.
just make it ` val errorCode: StrigaApiErrorCode = response.errorCode`, so this value could return the cached value

1sakov (approved):
minor comments.
lgtm

kabramovp2p (approved):

Davran331122 (approved):

PR #1802, opened 2023-06-08
PWN-8734 - Start integrating with sumsub sdk

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8734

### DIFF
--- a/app/src/main/java/org/p2p/wallet/striga/kyc/sdk/StrigaKycSdkListeners.kt
+++ b/app/src/main/java/org/p2p/wallet/striga/kyc/sdk/StrigaKycSdkListeners.kt
@@ -1,0 +1,53 @@
+package org.p2p.wallet.striga.kyc.sdk
+
+import com.sumsub.sns.core.SNSActionResult
+import com.sumsub.sns.core.data.listener.SNSEvent
+import com.sumsub.sns.core.data.model.SNSCompletionResult
+import com.sumsub.sns.core.data.model.SNSException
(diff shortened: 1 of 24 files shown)


app/src/main/java/org/p2p/wallet/striga/kyc/sdk/StrigaKycSdkListeners.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.sumsub.sns.core.data.model.SNSCompletionResult
+import com.sumsub.sns.core.data.model.SNSException
+import com.sumsub.sns.core.data.model.SNSSDKState
+
+/**
+ * Listener to know about errors that occur within the SDK
+ *
+ * [documentation](https://developers.sumsub.com/msdk/android/#on-sdk-errors)
+ */
+typealias StrigaSdkErrorListener = (error: SNSException) -> Unit
```
eduardmaximovich, line 14: I'm uncertain whether prefixing this with `Striga` is appropriate, given its actual association with SumSub. When we add another provider, this will likely be reused. As alternatives we could consider names such as `KycSdk*` or `SumSub*/Sumsub*`

gslevinkov: For another provider - we will have another service, that's for sure

Davran331122 (approved):

PR #1803, opened 2023-06-08
PWN-8853 - Change logic of country detection

Davran331122:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8853

### DIFF
--- a/app/src/main/java/org/p2p/wallet/auth/repository/CountryCodeInMemoryRepository.kt
+++ b/app/src/main/java/org/p2p/wallet/auth/repository/CountryCodeInMemoryRepository.kt
@@ -14,9 +14,7 @@ class CountryCodeInMemoryRepository(
     private val countryCodeHelper: CountryCodeXmlParser
 ) : CountryCodeRepository {
 
-    private val allCountryCodes: List<CountryCode> by lazy {
-        countryCodeHelper.parserCountryCodesFromXmlFile()
-    }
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/p2p/wallet/auth/repository/CountryCodeInMemoryRepository.kt
```diff
@@ -14,9 +14,7 @@ class CountryCodeInMemoryRepository(
     private val countryCodeHelper: CountryCodeXmlParser
 ) : CountryCodeRepository {
 
-    private val allCountryCodes: List<CountryCode> by lazy {
-        countryCodeHelper.parserCountryCodesFromXmlFile()
-    }
+    private val allCountryCodes: List<CountryCode> = countryCodeHelper.parserCountryCodesFromXmlFile()
```
gslevinkov, line 17: oh no, let's not do it on init instance

gradle.properties
```diff
@@ -1,4 +1,4 @@
-org.gradle.jvmargs=-Xmx2048m -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx5128m -Dfile.encoding=UTF-8
```
gslevinkov, line 1: wow, revert

gslevinkov (approved):

PR #1826, opened 2023-06-13
PWN-8349 - add logic to show proper data on screen when device is not same as in metadata

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8349

## Description of Work

PWN-8349 - add logic to show proper data on screen when device is not same as in metadata


app/src/main/java/org/p2p/wallet/settings/DeviceInfoHelper.kt
```diff
@@ -0,0 +1,10 @@
+package org.p2p.wallet.settings
+
+import android.os.Build
+
+object DeviceInfoHelper {
+
+    fun getCurrentDeviceName(): String {
+        return Build.MANUFACTURER + ' ' + Build.MODEL
```
gslevinkov, line 8: I suggest replacing all the occurrences.

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/settings/DeviceInfoHelper.kt
kabramovp2p: kek they already replaced =)

1sakov (approved):
minor comments

PR #1835, opened 2023-06-15
PWN-8920 - Fix preset data picker

Davran331122:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8920

### DIFF
--- a/app/src/main/java/org/p2p/wallet/striga/presetpicker/mapper/StrigaItemCellMapper.kt
+++ b/app/src/main/java/org/p2p/wallet/striga/presetpicker/mapper/StrigaItemCellMapper.kt
@@ -18,16 +18,13 @@ import org.p2p.wallet.striga.signup.model.StrigaSourceOfFunds
 class StrigaItemCellMapper {
 
     fun buildCellModels(items: List<StrigaPresetDataItem>, selectedItem: StrigaPresetDataItem?): List<AnyCellItem> {
-        val itemsWithoutSelected = items.filter { it != selectedItem }
         return buildList {
-            if (selectedItem != null) {
+            if (selectedItem?.getName()?.isNotEmpty() == true) {
                 this += buildHeaderCellItem(getSelectedItemHeaderTitle(selectedItem))
-                this += mapItemToCellItem(selectedItem)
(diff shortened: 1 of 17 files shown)


app/src/main/java/org/p2p/wallet/striga/presetpicker/mapper/StrigaItemCellMapper.kt
```diff
@@ -18,16 +18,13 @@ import org.p2p.wallet.striga.signup.model.StrigaSourceOfFunds
 class StrigaItemCellMapper {
 
     fun buildCellModels(items: List<StrigaPresetDataItem>, selectedItem: StrigaPresetDataItem?): List<AnyCellItem> {
-        val itemsWithoutSelected = items.filter { it != selectedItem }
         return buildList {
-            if (selectedItem != null) {
+            if (selectedItem?.getName()?.isNotEmpty() == true) {
```
gslevinkov, line 22: what does this mean? why we check name for emptyness?

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/striga/presetpicker/mapper/StrigaItemCellMapper.kt
Davran331122: I just write that, if name is empty, than selected item is null x D

Davran331122: I can make getName nullable

gslevinkov: so just check selectedItem != null?

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/striga/presetpicker/mapper/StrigaItemCellMapper.kt
Davran331122: it's not null anyway. but item.country could be null )

kabramovp2p (approved):

PR #1848, opened 2023-06-19
PWN-8938 - striga - mobile already used error handling

eduardmaximovich:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8938

## Description of Work

- Added mobile already user error handling while creating striga user

### DIFF
--- a/app/src/main/java/org/p2p/wallet/striga/model/StrigaDataLayerError.kt
+++ b/app/src/main/java/org/p2p/wallet/striga/model/StrigaDataLayerError.kt
@@ -9,14 +9,24 @@ sealed class StrigaDataLayerError(override val message: String) : Throwable() {
         body: String = cause.errorBodyOrNull().orEmpty()
     ) : StrigaDataLayerError("Striga API unavailable, code: ${cause.code()} body: $body")
 
-    open class ApiServiceError(
+    sealed class ApiServiceError(
         val response: StrigaApiErrorResponse,
     ) : StrigaDataLayerError("Striga API returned error: code=${response.errorCode.code} details=${response.details}") {
         val errorCode: StrigaApiErrorCode
             get() = response.errorCode
-    }
 
-    class PhoneNumberAlreadyUsedError(response: StrigaApiErrorResponse) : ApiServiceError(response)
+        class General(response: StrigaApiErrorResponse) : ApiServiceError(response)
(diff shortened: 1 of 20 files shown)


app/src/main/java/org/p2p/wallet/striga/model/StrigaDataLayerError.kt
```diff
@@ -9,13 +9,15 @@ sealed class StrigaDataLayerError(override val message: String) : Throwable() {
...
 
-    class ApiServiceError(
+    open class ApiServiceError(
         val response: StrigaApiErrorResponse,
     ) : StrigaDataLayerError("Striga API returned error: code=${response.errorCode.code} details=${response.details}") {
         val errorCode: StrigaApiErrorCode
             get() = response.errorCode
     }
 
+    class PhoneNumberAlreadyUsedError(response: StrigaApiErrorResponse) : ApiServiceError(response)
```
gslevinkov, line 19: we have StrigaApiErrorCode for this kind of things to remove inheritance
it doesn't work?

app/src/main/java/org/p2p/wallet/striga/user/repository/StrigaUserRepositoryMapper.kt
```diff
@@ -132,8 +132,8 @@ class StrigaUserRepositoryMapper {
             isEmailVerified = kycDetails.isEmailVerified,
             isMobileVerified = kycDetails.isMobileVerified,
             kycStatus = StrigaUserVerificationStatus.from(kycDetails.status),
-            rejectionDetails = kycDetails.details,
-            rejectionUserToAutoComments = kycDetails.rejectionComments.run { userComment to autoComment }
+            rejectionDetails = kycDetails.details ?: emptyList(),
```
gslevinkov, line 135: .orEmpty

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/striga/model/StrigaDataLayerError.kt
eduardmaximovich: Specializing errors helps to avoid duplicating code when extracting error codes. If you're not in favor of using inheritance in this case, I can create a separate error class with the same constructor instead.

kabramovp2p (approved):

PR #1849, opened 2023-06-19
PWN-8959 PWN-8953 PWN-8917 - Striga. Bugfixes

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8959 
https://p2pvalidator.atlassian.net/browse/PWN-8953
https://p2pvalidator.atlassian.net/browse/PWN-8917

Screenshot 2023-06-19 at 18 20 27

### DIFF
--- a/app/src/main/java/org/p2p/wallet/striga/ui/TopUpWalletPresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/striga/ui/TopUpWalletPresenter.kt
@@ -12,7 +12,6 @@ import org.p2p.wallet.common.InAppFeatureFlags
 import org.p2p.wallet.common.feature_toggles.toggles.remote.StrigaSignupEnabledFeatureToggle
 import org.p2p.wallet.common.mvp.BasePresenter
 import org.p2p.wallet.infrastructure.network.provider.SeedPhraseProvider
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/p2p/wallet/striga/ui/TopUpWalletPresenter.kt
```diff
@@ -29,19 +28,10 @@ class TopUpWalletPresenter(
...
-            return seedPhraseProvider.getUserSeedPhrase().provider == SeedPhraseSource.WEB_AUTH ||
-                appFeatureFlags.strigaSimulateWeb3Flag.featureValue
-        }
+        get() = seedPhraseProvider.isWeb3AuthUser || appFeatureFlags.strigaSimulateWeb3Flag.featureValue
 
     private val strigaBankTransferProgress = MutableStateFlow(false)
 
-    init {
-        launch {
-            loadMetadataIfNot()
```
gslevinkov, line 41: no need to load metadata on open topup 
it's needed only on Bank transfer, so if it's available - try to load when bank transfer is clicked. 
we can explicitly see loading and errors in this case without random "Something went wrong" when top up is opened

kabramovp2p (approved):

eduardmaximovich (approved):

PR #1856, opened 2023-06-21
PWN-9008 - Move crashlytics to separate package

Davran331122:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9008

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -89,6 +89,7 @@ dependencies {
     implementation(project(":ui-kit"))
     implementation(project(":core"))
     implementation(project(":ethereumkit"))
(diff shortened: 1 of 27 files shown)


app/build.gradle.kts
```diff
@@ -89,6 +89,7 @@ dependencies {
     implementation(project(":ui-kit"))
     implementation(project(":core"))
     implementation(project(":ethereumkit"))
+    implementation(project(":logger-crashlytics"))
```
gslevinkov, line 92: i thought this module is responsible only for crashlytics lib, but inside there is sentry too. 
maybe we can rename it? or don't extract it at all?

PR #1858, opened 2023-06-21
NO-TICKET - striga. kyc flow bugfixes

eduardmaximovich:
## Description of Work

- added debug button for removing striga from metadata
- added debug button for simulating striga rejected status
- added debug option for enabling striga email input
- loading striga user status at the key places
- fixed back button at the striga finish fragment
- fixed enabled/disabled state in UiKitEditText when setting value programmatically


app/src/main/java/org/p2p/wallet/debug/settings/DebugSettingsPresenter.kt
```diff
@@ -71,6 +75,32 @@ class DebugSettingsPresenter(
...
+                view?.showErrorMessage(e)
+            }
+        }
+    }
+
+    override fun onClickDetachStrigaUser() {
+        launch {
+            val metadata = metadataInteractor.currentMetadata
+            if (metadata == null) {
+                view?.showUiKitSnackBar("Metadata is not resolved. Unable to proceed")
```
gslevinkov, line 94: is not loaded would be better for UX

app/src/main/java/org/p2p/wallet/striga/kyc/ui/StrigaKycPresenter.kt
```diff
@@ -70,4 +79,15 @@ class StrigaKycPresenter(
...
     }
+
+    private fun updateUserStatus() {
+        appScope.launch(dispatchers.io) {
+            // update user status when kyc/start called
+            kotlin
+                .runCatching {
+                    strigaUserInteractor.loadAndSaveUserStatusData()
+                }
+                .onFailure { Timber.e(it, "Unable to load striga user status") }
```
gslevinkov, line 90:
```suggestion
            kotlin.runCatching { strigaUserInteractor.loadAndSaveUserStatusData() }
                .onFailure { Timber.e(it, "Unable to load striga user status") }
```

app/src/main/java/org/p2p/wallet/striga/signup/interactor/StrigaSignupInteractor.kt
```diff
@@ -156,6 +156,10 @@ class StrigaSignupInteractor(
             }
         }
 
+        // not critical, we can ignore errors from this call
+        kotlin.runCatching {
+            userInteractor.loadAndSaveUserStatusData().unwrap()
```
gslevinkov, line 161: you can create a method ignore or just call userInteractor.loadAndSaveUserStatusData()

until you call unwrap - no exception will be thrown

gslevinkov (approved):

PR #1871, opened 2023-06-22
PWN-8969 - striga. sms errors handling

eduardmaximovich:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8968

## Description of Work

- Moved sms send logic to StrigaSmsInputPresenter
- Added limits internal timer to show "exceeded" errors without api call
- Removed sending verification sms after user creation as it is unnecessary
- Removed updating striga user status right after user creation
- Added `DateTimeUtils.isTimestampExpired(timeMs: Long, expireDurationMs: Long): Boolean` which checks whether a timer is already completed


app/src/main/java/org/p2p/wallet/debug/settings/DebugSettingsInteractor.kt
```diff
@@ -0,0 +1,25 @@
+package org.p2p.wallet.debug.settings
+
+import org.p2p.wallet.auth.interactor.MetadataInteractor
+import org.p2p.wallet.striga.kyc.ui.StrigaKycInteractor
+import org.p2p.wallet.striga.user.StrigaStorageContract
+
+class DebugSettingsInteractor(
```
eduardmaximovich, line 7: extracted logic to the interactor, the presenter is already doing too much work

app/src/main/java/org/p2p/wallet/striga/model/StrigaDataLayerResult.kt
```diff
@@ -1,8 +1,8 @@
 package org.p2p.wallet.striga.model
 
-sealed interface StrigaDataLayerResult<out T> {
-    data class Success<T>(val value: T) : StrigaDataLayerResult<T>
-    class Failure<T>(val error: StrigaDataLayerError) : StrigaDataLayerResult<T>
+sealed class StrigaDataLayerResult<out T> {
```
eduardmaximovich, line 3: changed to `class` because cannot use inline functions within non-final interfaces

app/src/main/java/org/p2p/wallet/striga/signup/interactor/StrigaSignupInteractor.kt
```diff
@@ -156,10 +156,6 @@ class StrigaSignupInteractor(
                 throw result.error
             }
         }
-
-        // not critical, we can ignore errors from this call
-        userInteractor.loadAndSaveUserStatusData()
-        userInteractor.resendSmsForVerifyPhoneNumber().unwrap()
```
eduardmaximovich, line 162: not needed anymore

app/src/main/java/org/p2p/wallet/striga/signup/ui/StrigaSignUpSecondStepFragment.kt
```diff
@@ -100,6 +102,9 @@ class StrigaSignUpSecondStepFragment :
         replaceFragment(
             SmsInputFactory.create(
                 type = SmsInputFactory.Type.Striga,
+                args = bundleOf(
+                    StrigaSmsInputFragment.ARG_RESEND_SMS_ON_INIT to false
+                ),
```
eduardmaximovich, line 107: do not send sms right after user creation, striga already do this for us without calling `resend-sms`

app/src/main/java/org/p2p/wallet/utils/DateTimeUtils.kt
```diff
@@ -156,6 +156,10 @@ object DateTimeUtils {
         return String.format(Locale.US, "%02d:%02d", m, s)
     }
 
+    fun checkTimerIsOver(timeMs: Long, expireDurationMs: Long): Boolean {
```
gslevinkov, line 159: well, I don't think DateTimeUtils should know about any timer staff :)) 
maybe `isTimestampExpired`

eduardmaximovich: that's what I wanted! - isTimestampExpired

kabramovp2p (approved):

gslevinkov (approved):

PR #1872, opened 2023-06-23
PWN-9001 - Swap refresh routes interval via feature toggle

1sakov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9001

## Description of Work

## Screenshots (Optional)

### DIFF
--- a/app/src/main/java/org/p2p/wallet/common/feature_toggles/remote_config/AppFirebaseRemoteConfig.kt
+++ b/app/src/main/java/org/p2p/wallet/common/feature_toggles/remote_config/AppFirebaseRemoteConfig.kt
@@ -106,6 +106,13 @@ class AppFirebaseRemoteConfig(
         return remoteConfig.getString(toggleKey).toIntOrNull()
     }
 
+    override fun getLong(toggleKey: String): Long? {
+        if (isFetchFailed) {
+            return null
+        }
+        return remoteConfig.getString(toggleKey).toLongOrNull()
+    }
+
     override fun getFloat(toggleKey: String): Float? {
         if (isFetchFailed) {
             return null
--- a/app/src/main/java/org/p2p/wallet/common/feature_toggles/toggles/remote/RemoteFeatureToggle.kt
+++ b/app/src/main/java/org/p2p/wallet/common/feature_toggles/toggles/remote/RemoteFeatureToggle.kt
@@ -1,7 +1,7 @@
 package org.p2p.wallet.common.feature_toggles.toggles.remote
 
-import org.p2p.wallet.common.feature_toggles.remote_config.RemoteConfigValuesProvider
 import timber.log.Timber
+import org.p2p.wallet.common.feature_toggles.remote_config.RemoteConfigValuesProvider
 
 sealed class RemoteFeatureToggle<ValueType> {
     /**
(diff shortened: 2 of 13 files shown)


app/src/main/java/org/p2p/wallet/common/feature_toggles/toggles/remote/RemoteFeatureToggle.kt
```diff
@@ -39,6 +39,16 @@ abstract class IntFeatureToggle(
         }
 }
 
+abstract class LongFeatureToggle(
+    private val valuesProvider: RemoteConfigValuesProvider
+) : RemoteFeatureToggle<Long>() {
+    override val value: Long
+        get() = valuesProvider.getLong(featureKey) ?: kotlin.run {
+            Timber.tag("IntFeatureToggle").i("No value found for $featureKey; using defaults = $defaultValue")
```
gslevinkov, line 47: IntFeatureToggle -> LongFeatureToggle

app/src/main/java/org/p2p/wallet/common/feature_toggles/toggles/remote/SwapRoutesRefreshFeatureToggle.kt
```diff
@@ -0,0 +1,13 @@
...
+import org.p2p.wallet.common.feature_toggles.remote_config.RemoteConfigValuesProvider
+
+private const val DEFAULT_INTERVAL_IN_MILLIS = 20_000L
+
+class SwapRoutesRefreshFeatureToggle(
+    valuesProvider: RemoteConfigValuesProvider
+) : LongFeatureToggle(valuesProvider) {
+    override val featureKey: String = "swap_route_refresh"
+    override val featureDescription: String = "The interval for refreshing routes"
+    override val defaultValue: Long = DEFAULT_INTERVAL_IN_MILLIS
```
gslevinkov, line 12: what if you add here valueDuration: Duration and use it like 
val valueDuration:Duration = value.milliseconds

so we won't guess what time unit is inside the actual value

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/common/feature_toggles/remote_config/AppFirebaseRemoteConfig.kt
```diff
@@ -106,6 +106,13 @@ class AppFirebaseRemoteConfig(
         return remoteConfig.getString(toggleKey).toIntOrNull()
     }
 
+    override fun getLong(toggleKey: String): Long? {
```
eduardmaximovich, line 109: Perhaps for clarity, all these `get*` functions should be separated into two distinct ones:
1. `getLongOrNull(): Long?`
2. `getLong(key, default: Long): Long`

eduardmaximovich (approved):

app/src/main/java/org/p2p/wallet/common/feature_toggles/remote_config/AppFirebaseRemoteConfig.kt
gslevinkov: it's a inside impl of FirebaseRemoteConfig
we pass default values inside the feature toggles, so i think there no need for multiple methods

PR #1882, opened 2023-06-27
PWN-9042 - reset torus timestamp on error to prevent issue

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9042

## Description of Work

PWN-9042 - reset torus timestamp on error to prevent issue


1sakov (approved):

app/src/main/java/org/p2p/wallet/auth/repository/RestoreFlowDataLocalRepository.kt
```diff
@@ -103,4 +103,8 @@ class RestoreFlowDataLocalRepository(signUpDetailsStorage: UserSignUpDetailsStor
     fun isTorusKeyValid(): Boolean {
         return DateTimeUtils.getCurrentTimestampInSeconds() - torusKeyTimestamp < 15.minutes.inWholeSeconds
     }
+
+    fun resetTorusTimestamp() {
```
gslevinkov, line 107: can you put some comment on why do we need to reset torus key timestamp?

gslevinkov (approved):

PR #1883, opened 2023-06-27
PWN-8904 - Striga API. get_user_wallets

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8904

### DIFF
--- a/app/src/main/java/org/p2p/wallet/striga/wallet/repository/StrigaWalletRemoteRepository.kt
+++ b/app/src/main/java/org/p2p/wallet/striga/wallet/repository/StrigaWalletRemoteRepository.kt
@@ -112,7 +112,7 @@ class StrigaWalletRemoteRepository(
         }
     }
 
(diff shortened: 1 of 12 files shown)


1sakov (approved):

app/src/main/java/org/p2p/wallet/striga/wallet/repository/StrigaWalletRemoteRepository.kt
```diff
@@ -105,4 +111,27 @@ class StrigaWalletRemoteRepository(
             )
         }
     }
+
+    override suspend fun getUserWallet(): StrigaDataLayerResult<StrigaUserWallet?> {
```
eduardmaximovich, line 115: let's have this function return a non-null type and throw an exception if it can't find any wallets, that would be a pretty weird case, I reckon

eduardmaximovich (approved):

PR #1884, opened 2023-06-27
PWN-8904 - Striga API. get_estimated_fees

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8904


app/src/main/java/org/p2p/wallet/striga/wallet/api/response/StrigaOnchainWithdrawalFeeResponse.kt
```diff
@@ -0,0 +1,20 @@
+package org.p2p.wallet.striga.wallet.api.response
+
+import com.google.gson.annotations.SerializedName
+
+data class StrigaOnchainWithdrawalFeeResponse(
```
gslevinkov, line 5: is reused across two api requests

eduardmaximovich (approved):

PR #1890, opened 2023-06-28
PWN-8973 - striga. signup otp timer

eduardmaximovich:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8973

## Description of Work

- added timers to striga signup otp input


app/src/main/java/org/p2p/wallet/auth/AuthModule.kt
```diff
@@ -145,8 +148,40 @@ object AuthModule {
                 durationTracker = get()
             )
         }
+        single(named(SMS_TIMER_ONBOARDING)) {
+            SmsInputTimer(get())
+        }
```
gslevinkov, line 153:
```suggestion
        singleOf(::SmsInputTimer) { 
            named(SMS_TIMER_ONBOARDING)
       }
```

and I would rather use enum classes instead of strings

app/src/main/java/org/p2p/wallet/smsinput/striga/StrigaSmsInputInteractor.kt
```diff
@@ -74,22 +78,31 @@ class StrigaSmsInputInteractor(
...
+                    }
+                    StrigaApiErrorCode.EXCEEDED_DAILY_RESEND_SMS_LIMIT -> {
+                        setExceededResendAttempts()
+                    }
+                    else -> Unit
+                }
+            }
+            .onSuccess {
+                smsInputTimer.startSmsInputTimerFlow()
+            }
```
gslevinkov, line 103: ?: block on the left can be extracted to a  method

app/src/main/java/org/p2p/wallet/smsinput/striga/StrigaSmsInputPresenter.kt
```diff
@@ -68,6 +52,55 @@ class StrigaSmsInputPresenter(
...
+                    handleError(result.error)
+                }
+            }
+            view?.renderButtonLoading(isLoading = false)
+        }
+    }
+
+    private fun checkForExceededLimits() {
+        interactor.getExceededLimitsErrorIfPresent()?.let {
+            require(it.error is StrigaDataLayerError.ApiServiceError) {
```
gslevinkov, line 72: i would rather not throw exception in the runtime. maybe if is enough here?

app/src/main/java/org/p2p/wallet/striga/di/StrigaSignupModule.kt
```diff
@@ -73,13 +75,25 @@ object StrigaSignupModule : InjectionModule {
...
         factoryOf(::StrigaSignupInteractor)
         factoryOf(::StrigaUserInteractor)
-        factoryOf(::StrigaSmsInputInteractor)
-        factory(named(SmsInputFactory.Type.Striga.name)) { (resendSmsOnLaunch: Boolean) ->
-            StrigaSmsInputPresenter(
-                resendSmsOnLaunch = resendSmsOnLaunch,
-                interactor = get()
-            )
+
+        factory(named(SmsInputFactory.Type.Striga.name)) {
```
gslevinkov, line 79: same
can be rewritten using Koin DSL

gslevinkov (approved):

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/auth/AuthModule.kt
eduardmaximovich: fixed except enum, next time with on-ramp flow

app/src/main/java/org/p2p/wallet/smsinput/striga/StrigaSmsInputInteractor.kt
```diff
@@ -25,7 +27,19 @@ class StrigaSmsInputInteractor(
     private val phoneCodeRepository: CountryCodeRepository,
     private val inAppFeatureFlags: InAppFeatureFlags,
     private val strigaStorage: StrigaStorageContract,
+    private val smsInputTimer: SmsInputTimer,
 ) {
+    val timer: Flow<Int> get() = smsInputTimer.smsInputTimerFlow
+    val isTimerNotActive: Boolean get() = !smsInputTimer.isTimerActive
```
eduardmaximovich, line 33: it is used only in negative checking, so I decided to inverse the value

app/src/main/java/org/p2p/wallet/smsinput/striga/StrigaSmsInputInteractor.kt
```diff
@@ -74,22 +88,29 @@ class StrigaSmsInputInteractor(
...
-                            else -> Unit
-                        }
-                    }
+                null
             }
         }
     }
 
+    suspend fun resendSms(): StrigaDataLayerResult<Unit> {
+        return getExceededLimitsErrorIfPresent() ?: strigaUserRepository.resendSmsForVerifyPhoneNumber()
```
gslevinkov, line 97: and again, i suggest extracting the whole block from the left of ?: to a method

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/smsinput/striga/StrigaSmsInputInteractor.kt
eduardmaximovich: @gslevinkov Do you suggest moving logic from getExceededLimitsErrorIfPresent to strigaUserRepository.resendSmsForVerifyPhoneNumber ? 
This function intentionally was separated because it is also used in the presenter to get this error without an api call. This means we would have to call resendSms even when it's not necessary

PR #1891, opened 2023-06-28
PWN-8755 - move metadata to account storage as SecureStorageContract.kt cleared after logout

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8755

## Description of Work

PWN-8755 - move metadata to account storage as SecureStorageContract.kt cleared after logout


app/src/main/java/org/p2p/wallet/auth/interactor/MetadataInteractor.kt
```diff
@@ -90,8 +90,8 @@ class MetadataInteractor(
     private fun saveMetadataToStorage(metadata: GatewayOnboardingMetadata?) {
         // TODO PWN-8771 - implement database for metadata
         val ethAddress = getEthereumPublicKey().orEmpty()
-        secureStorageContract.saveObject(
-            key = SecureStorageContract.Key.KEY_ONBOARDING_METADATA.withCustomKey(ethAddress),
+        accountStorage.saveObject(
```
gslevinkov, line 93: sorry, but no. 
You need a migration code for all of this.
Extract everything to MetadataProvider and rule there

gslevinkov (approved):

PR #1900, opened 2023-06-29
NO-TICKET - Updated libs and moved all them to [Dependencies.kt]

1sakov:
https://kotlinlang.org/docs/whatsnew18.html

https://developer.android.com/jetpack/androidx/releases/browser#1.5.0

https://github.com/bumptech/glide/releases

https://developer.android.com/jetpack/androidx/releases/ads#1.0.0-alpha05

https://developer.android.com/jetpack/androidx/releases/core#core-splashscreen-1.0.1

https://developer.android.com/jetpack/androidx/releases/core#version_10_2


gslevinkov (approved):

app/src/main/java/org/p2p/wallet/history/ui/token/TokenHistoryFragment.kt
```diff
@@ -170,6 +170,9 @@ class TokenHistoryFragment :
                 sellAnalytics.logTokenScreenActionClicked()
                 replaceFragment(SellPayloadFragment.create())
             }
+            else -> {
+                // TODO: WHAT SHOULD BE HERE?
+            }
```
eduardmaximovich, line 175: just insert Unit, it's totally ok

```suggestion
            else -> Unit
```

buildSrc/src/main/java/Versions.kt
```diff
@@ -25,5 +25,7 @@ object Versions {
         return "key-app-$VERSION_NAME.$VERSION_PATCH"
     }
 
-    const val kotlinVersion = "1.7.20"
+    const val kotlinVersion = "1.8.10"
```
eduardmaximovich, line 28: 🔥

eduardmaximovich (approved):

PR #1903, opened 2023-06-29
PWN-8908 - striga. claim otp + sms logic refactoring

eduardmaximovich:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8908
https://p2pvalidator.atlassian.net/browse/PWN-8898
https://p2pvalidator.atlassian.net/browse/PWN-8897

## Description of Work

- striga claim (on-ramp) OTP + api methods + a bit refactored logic for reusability


app/src/main/java/org/p2p/wallet/common/NavigationStrategy.kt
```diff
@@ -17,17 +17,18 @@ sealed class NavigationStrategy : Parcelable {
...
+            return nextDestinationClass?.newInstance()?.apply {
                 arguments = nextDestinationArgs
             }
         }
     }
 
     object Replace : NavigationStrategy() {
-        override fun execute(sourceFragment: Fragment, destinationFragment: Fragment) {
+        override fun execute(sourceFragment: Fragment, destinationFragment: Fragment?) {
+            require(destinationFragment != null) { "Destination fragment must not be null" }
```
gslevinkov, line 31: requireNotNull

app/src/main/java/org/p2p/wallet/striga/sms/onramp/StrigaOnRampSmsInputFragment.kt
```diff
@@ -0,0 +1,70 @@
...
+class StrigaOnRampSmsInputFragment : BaseSmsInputFragment() {
+
+    companion object {
+        const val ARG_TITLE_AMOUNT = "ARG_TITLE_AMOUNT"
+        const val ARG_CHALLENGE_ID = "ARG_CHALLENGE_ID"
+        val REQUEST_KEY: String = StrigaOnRampSmsInputFragment::class.java.name
+    }
+
+    private val titleAmount: String by args(ARG_TITLE_AMOUNT)
+    private val challengeId: StrigaWithdrawalChallengeId by args(ARG_CHALLENGE_ID)
```
gslevinkov, line 33: wow
that's new
what is a challenge?

app/src/main/java/org/p2p/wallet/striga/sms/onramp/StrigaOnRampSmsStorage.kt
```diff
@@ -0,0 +1,21 @@
+package org.p2p.wallet.striga.sms.onramp
+
+import org.p2p.wallet.striga.sms.StrigaSmsStorageContract
+import org.p2p.wallet.striga.user.StrigaStorageContract
+
+class StrigaOnRampSmsStorage(
+    private val storage: StrigaStorageContract
+) : StrigaSmsStorageContract {
+
+    override var exceededVerificationAttemptsMillis: Long
```
gslevinkov, line 10: you can do Duration instead of Long

or replace long with EpochSinceMillis

app/src/main/java/org/p2p/wallet/striga/sms/onramp/StrigaOnRampSmsStorage.kt
```diff
@@ -0,0 +1,21 @@
+package org.p2p.wallet.striga.sms.onramp
+
+import org.p2p.wallet.striga.sms.StrigaSmsStorageContract
+import org.p2p.wallet.striga.user.StrigaStorageContract
+
+class StrigaOnRampSmsStorage(
```
gslevinkov, line 6: do we really need a proxy class for these fields? 
maybe we can put them inside StrigaStorageContract and leave only two of them? 
i think logic is same for them

app/src/main/java/org/p2p/wallet/striga/wallet/repository/StrigaWalletRemoteRepository.kt
```diff
@@ -154,4 +161,62 @@ class StrigaWalletRemoteRepository(
...
+                error = error,
+                default = StrigaDataLayerError.InternalError(error)
+            )
+        }
+    }
+
+    /**
+     * This is a temporary function, until we have a proper way to get the ip address
+     */
+    private fun getLocalIpAddress(): String {
```
gslevinkov, line 207: hey, let's go with a separate util class like LocalIpAddressProvider so we can reuse that in case of smth

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/striga/sms/onramp/StrigaOnRampSmsInputFragment.kt
eduardmaximovich: this model was introduced with initiate/withdrawal api call, this id is just a sort of internal transaction-id in fact

app/src/main/java/org/p2p/wallet/striga/sms/onramp/StrigaOnRampSmsStorage.kt
eduardmaximovich: You're right. I just realized, that it actually doesn't make sense to even have 2 different timers for both screens, because they don't overlap each other at all, we can use the same fields for different blocking timers.

app/src/main/java/org/p2p/wallet/striga/sms/onramp/StrigaOnRampSmsInputFragment.kt
gslevinkov: so maybe we can call it just StrigaWithdrawalTransactionId or StrigaWithdrawalId?

app/src/main/java/org/p2p/wallet/striga/wallet/repository/StrigaWalletRemoteRepository.kt
eduardmaximovich: I will move this to the StrigaIpAddressProvider, as we will obtain the actual external IP address from our striga backend

app/src/main/java/org/p2p/wallet/striga/sms/onramp/StrigaOnRampSmsStorage.kt
eduardmaximovich: replaced with EpochSinceMillis, because logic for Duration needs to be fully retested

app/src/main/java/org/p2p/wallet/striga/sms/onramp/StrigaOnRampSmsInputFragment.kt
eduardmaximovich: i'm not sure that is a good idea to change it to `transaction id` because the `challenge` actually means that the system is asking for a user authentication and the `challengeId` represents some temporary key to use in this authentication

app/src/main/java/org/p2p/wallet/home/ui/main/HomeFragment.kt
```diff
@@ -309,11 +311,15 @@ class HomeFragment :
...
-        } else {
-            strigaKycFragmentFactory.kycFragment()
-        }.also(::replaceFragment)
+//        if (status == StrigaKycStatusBanner.VERIFICATION_DONE) {
+//            StrigaUserIbanDetailsFragment.create()
+//        } else {
+//            strigaKycFragmentFactory.kycFragment()
+//        }.also(::replaceFragment)
+        val f = strigaKycFragmentFactory.claimOtpFragment("$100", StrigaWithdrawalChallengeId("a"))
+        replaceFragmentForResult(f, "claim_otp", onResult = { _, _ ->
```
gslevinkov, line 320: ?

app/src/main/java/org/p2p/wallet/striga/common/StrigaIpAddressProvider.kt
```diff
@@ -0,0 +1,28 @@
...
+                for (inetAddress in networkInterface.inetAddresses) {
+                    if (!inetAddress.isLoopbackAddress && inetAddress is Inet4Address) {
+                        return inetAddress.getHostAddress() ?: defaultOne
+                    }
+                }
+            }
+        } catch (e: SocketException) {
+            Timber.e(e, "Unable to retrieve ip address")
+        }
+        return defaultOne
```
gslevinkov, line 26:
```suggestion
val defaultOne = "127.0.0.1"
        return try {
            NetworkInterface.getNetworkInterfaces().toList()
                .flatMap { it.inetAddresses.toList() }
                .firstOrNull { it is Inet4Address && !it.isLoopbackAddress  }
                ?.hostAddress
                ?: defaultOne
        } catch (e: SocketException) {
            Timber.e(e, "Unable to retrieve ip address")
            defaultOne
        }
```

gslevinkov: not forcing anything, just a variant of your impl

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/home/ui/main/HomeFragment.kt
eduardmaximovich: bah, removed

app/src/main/java/org/p2p/wallet/striga/sms/onramp/StrigaOnRampSmsInputFragment.kt
```diff
@@ -0,0 +1,70 @@
...
+        setFragmentResult(REQUEST_KEY, Bundle.EMPTY)
+        super.navigateNext()
+    }
+
+    override fun initView(userPhoneNumber: PhoneNumber) {
+        binding.textViewDescription.text =
+            getString(R.string.striga_onramp_sms_input_description, userPhoneNumber.formattedValue)
+    }
+
+    override fun navigateToSmsInputBlocked(timerLeftTime: Long) = Unit
```
Davran331122, line 59: I think better to have some enum class, and incapsulate in to one function

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/striga/sms/onramp/StrigaOnRampSmsInputPresenter.kt
```diff
@@ -0,0 +1,143 @@
...
+                    )
+                )
+                if (secondsBeforeResend == 0) {
+                    view?.renderSmsTimerState(SmsInputContract.Presenter.SmsInputTimerState.ResendSmsReady)
+                }
+            }
+        }
+    }
+
+    private fun isSmsCodeFormatValid(smsCode: String): Boolean = smsCode.length == 6
```
Davran331122, line 112: you can move 6 to const

Davran331122 (approved):

Davran331122 (approved):

app/src/main/java/org/p2p/wallet/striga/sms/onramp/StrigaOnRampSmsInputFragment.kt
eduardmaximovich: it definitely requires some refactoring, but for now introducing new navigation logic would require to test again all the 3 screens, it's out of the tasks scope

PR #1904, opened 2023-06-29
PWN-9119 - Create Database for saving token prices

Davran331122:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9119


core/build.gradle.kts
```diff
@@ -32,6 +33,9 @@ dependencies {
     api(Dependencies.timber)
     implementation(Dependencies.glide)
 
+    api(Dependencies.roomRuntime)
+    api(Dependencies.roomKtx)
```
gslevinkov, line 37: I don't think it's needed everywhere, I'd put them where they are needed.

token-service/src/main/java/org/p2p/token/service/TokenServiceModule.kt
```diff
@@ -61,5 +63,7 @@ object TokenServiceModule : InjectionModule {
         )}
         factoryOf(::TokenServiceEventPublisher)
         singleOf(::TokenServiceEventManager)
+
+        loadKoinModules(TokenServiceDatabaseModule.create())
```
gslevinkov, line 67: why load? 
you can use includes

token-service/src/main/java/org/p2p/token/service/database/TokenServiceDatabaseModule.kt
```diff
@@ -0,0 +1,24 @@
...
+import org.koin.dsl.module
+import java.util.concurrent.Executors
+import org.p2p.core.common.di.InjectionModule
+
+internal const val TOKEN_SERVICE_DB_QUALIFIER = "TOKEN_SERVICE_DB_QUALIFIER"
+
+internal object TokenServiceDatabaseModule : InjectionModule {
+
+    override fun create(): Module = module {
+        single(named(TOKEN_SERVICE_DB_QUALIFIER)) {
```
gslevinkov, line 16: and why do you need named single here? is there multiple TokenServiceDatabase?

token-service/src/main/java/org/p2p/token/service/database/converter/TokenServiceRoomConverter.kt
```diff
@@ -0,0 +1,12 @@
+package org.p2p.token.service.database.converter
+
+import androidx.room.TypeConverter
+import org.p2p.token.service.model.TokenServiceNetwork
+
+internal object  TokenServiceRoomConverter {
+    @TypeConverter
+    fun tokenServiceNetworkToString(value: TokenServiceNetwork?) = value?.networkName
```
gslevinkov, line 8: explicit return type

token-service/src/main/java/org/p2p/token/service/database/converter/TokenServiceRoomConverter.kt
```diff
@@ -0,0 +1,12 @@
...
+
+    @TypeConverter
+    fun stringToTokenServiceNetwork(value: String) = TokenServiceNetwork.valueOf(value)
```
gslevinkov, line 11: tokenServiceNetworkToString (TokenServiceNetwork?) -> String?

stringToTokenServiceNetwork doesn't accept any nullable value which is strange, so I think TypeConverter won't be used

token-service/src/main/java/org/p2p/token/service/database/entity/TokenServicePriceEntity.kt
```diff
@@ -0,0 +1,27 @@
...
+    @PrimaryKey
+    @ColumnInfo(name = COLUMN_TOKEN_ADDRESS)
+    val tokenAddress: String,
+    @ColumnInfo(name = COLUMN_TOKEN_PRICE)
+    val price: BigDecimal?,
+    @ColumnInfo(name = COLUMN_TOKEN_NETWORK_CHAIN)
+    val network: TokenServiceNetwork
+) {
+    companion object {
+        const val TABLE_NAME = "token_price_table"
```
gslevinkov, line 22: `token_prices` would be sufficient

token-service/src/main/java/org/p2p/token/service/database/entity/TokenServicePriceEntity.kt
```diff
@@ -0,0 +1,27 @@
...
+
+@Entity(
+    tableName = TokenServicePriceEntity.TABLE_NAME
+)
+data class TokenServicePriceEntity(
+    @PrimaryKey
+    @ColumnInfo(name = COLUMN_TOKEN_ADDRESS)
+    val tokenAddress: String,
+    @ColumnInfo(name = COLUMN_TOKEN_PRICE)
+    val price: BigDecimal?,
```
gslevinkov, line 17: if we want to support multicurrency in future, it's better to think about it now

also, i don't think we need to write entity to DB if price is null. 
just check before saving and save only that prices that are not null

token-service/src/main/java/org/p2p/token/service/database/mapper/TokenServicDatabaseMapper.kt
```diff
@@ -0,0 +1,4 @@
+package org.p2p.token.service.database.mapper
+
+class TokenServicDatabaseMapper {
```
gslevinkov, line 3: niiceee

gslevinkov (changes requested):

token-service/src/main/java/org/p2p/token/service/database/TokenServiceDatabaseModule.kt
Davran331122: just to be sure, that everything is OK )

app/src/main/java/org/p2p/wallet/App.kt
```diff
@@ -90,7 +90,7 @@ class App : Application(), Configuration.Provider {
             // crashes when using level != Level.Error
             // do NOT use other than ERROR before bumping to Koin 3.1.6
             // FIXME
-            androidLogger(level = Level.DEBUG)
+            androidLogger(level = Level.INFO)
```
Davran331122, line 93: Revert this

app/src/main/java/org/p2p/wallet/home/ui/main/HomePresenter.kt
```diff
@@ -463,27 +464,29 @@ class HomePresenter(
...
+            state = state.copy(
+                tokens = userTokens,
+                ethTokens = ethTokens,
+                username = homeInteractor.getUsername(),
+            )
+            val isAccountEmpty = userTokens.all(Token.Active::isZero) && ethTokens.isEmpty()
+            when {
+                isAccountEmpty -> {
+                    view?.showEmptyState(isEmpty = true)
+                    handleEmptyAccount()
```
eduardmaximovich, line 478: You can call `launch { handleEmptyAccount() }` instead of wrapping the entire function in `launch` call as there are no other suspend functions being called.
Alternatively, if the data is strongly related, you could add `suspend` keyword to the function, and let other functions decide in which coroutine it should run

token-service/src/main/java/org/p2p/token/service/model/TokenServiceNetwork.kt
```diff
@@ -1,6 +1,13 @@
...
-enum class TokenServiceNetwork {
-    SOLANA,
-    ETHEREUM
+enum class TokenServiceNetwork(val networkName: String) {
+    SOLANA("solana"),
+    ETHEREUM("ethereum");
+
+    companion object {
+        fun getValueOf(stringValue: String): TokenServiceNetwork {
+            return TokenServiceNetwork.values().firstOrNull { it.networkName == stringValue }
```
eduardmaximovich, line 9: I would add case-insensitive comparison just in case
```suggestion
            TokenServiceNetwork.values().firstOrNull { it.networkName.equals(stringValue, ignoreCase = true) }
```

token-service/src/main/java/org/p2p/token/service/repository/price/TokenPriceDatabaseRepository.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.p2p.token.service.model.TokenServicePrice
+
+internal class TokenPriceDatabaseRepository(
+    private val tokenPriceDao: TokenPriceDao,
+    private val converter: TokenServiceDatabaseMapper,
+    coroutineDispatchers: CoroutineDispatchers
+) : TokenPriceLocalRepository, CoroutineScope {
+
+    override val coroutineContext: CoroutineContext = coroutineDispatchers.io
+    override fun setTokensPrice(prices: List<TokenServicePrice>) {
```
eduardmaximovich, line 21: why isn't it a suspend function?

token-service/src/main/java/org/p2p/token/service/database/TokenPriceDao.kt
```diff
@@ -0,0 +1,35 @@
...
+    fun getAllTokensPricesFlow(): Flow<TokenPriceEntity>
+
+    @Query("SELECT * FROM token_price_table WHERE token_address =:tokenAddress")
+    suspend fun getTokenPriceByAddress(tokenAddress: String): TokenPriceEntity?
+
+    @Query("SELECT * FROM token_price_table WHERE token_network_chain =:networkChain")
+    suspend fun getTokensRatesByNetwork(networkChain: String): List<TokenPriceEntity>
+
+    @Query("SELECT * FROM token_price_table WHERE token_network_chain =:networkChain ")
+    fun getTokensRatesByNetworkFlow(networkChain: String): Flow<List<TokenPriceEntity>>
```
eduardmaximovich, line 34: please, add space between `=` and `:{arg}`

token-service/src/main/java/org/p2p/token/service/model/TokenServicePrice.kt
```diff
@@ -5,8 +5,10 @@ import org.p2p.core.utils.scaleShort
...
 class TokenServicePrice(
     val address: String,
-    val price: BigDecimal
+    val rate: TokenRate?,
+    val network: TokenServiceNetwork
 ) {
-    fun getScaledValue(): BigDecimal = price.scaleShort()
+    fun getScaledUsdRate(): BigDecimal? = rate?.usd?.scaleShort()
 
+    fun getUsdRate(): BigDecimal? = rate?.usd
```
gslevinkov, line 13: can you make it as a property?
it's more Kotlin-style

token-service/src/main/java/org/p2p/token/service/repository/TokenServiceRepositoryImpl.kt
```diff
@@ -42,22 +42,21 @@ class TokenServiceRepositoryImpl(
         }
     }
 
-    override fun getTokenPricesFlow(networkChain: TokenServiceNetwork): StateFlow<Map<String, TokenServicePrice>> =
+    override suspend fun getTokenPricesFlow(networkChain: TokenServiceNetwork): Flow<List<TokenServicePrice>> =
```
gslevinkov, line 45: observeTokenPrices

token-service/src/main/java/org/p2p/token/service/repository/price/TokenPriceDatabaseRepository.kt
gslevinkov, line 21: saveTokensPrices

token-service/src/main/java/org/p2p/token/service/repository/price/TokenPriceDatabaseRepository.kt
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    override suspend fun findTokenPriceByAddress(
+        address: String
+    ): TokenServicePrice? {
+        val found = tokenPriceDao.getTokenPriceByAddress(address)
+        return found?.let { converter.fromEntity(it) }
+    }
+
+    override suspend fun attachToTokensPrice(networkChain: TokenServiceNetwork): Flow<List<TokenServicePrice>> {
```
gslevinkov, line 35: attach is used in MVP context, better to use observeTokensPrices

gslevinkov (approved):

token-service/src/main/java/org/p2p/token/service/repository/price/TokenPriceDatabaseRepository.kt
eduardmaximovich: Nonetheless, this function is essentially a suspend function because it's entirely wrapped in the launch block.
If you want to maintain specific dispatcher for this specific db operation you can just use `withContext(YOUR_DB_DISPATCHER)`.

Davran331122: I meant, that in this way we make it async, and don't suspend launch thread from presenter

eduardmaximovich (approved):

app/src/main/java/org/p2p/wallet/restore/ui/derivable/DerivableAccountsPresenter.kt
```diff
@@ -41,12 +41,11 @@ class DerivableAccountsPresenter(
     private fun loadSolRate() {
         launch {
             try {
-                val tokenMint = Constants.SOL_MINT
+                val tokenMint = Constants.WRAPPED_SOL_MINT
                 val solRate = tokenServiceRepository.fetchTokenPriceByAddress(
                     networkChain = TokenServiceNetwork.SOLANA,
                     tokenAddress = tokenMint
```
eduardmaximovich, line 47: I think it should be simplified for clarity (the entire body of `try`)
```suggestion
                tokenServiceRepository.fetchTokenPriceByAddress(
                    networkChain = TokenServiceNetwork.SOLANA,
                    tokenAddress = tokenMint
                )?.also {
                    solRate = it
                    allAccounts = allAccounts.updateWithTotalInUsd(it).toMutableList()
                    filterAccountsByPathAndShow(path)
                }
```

PR #1905, opened 2023-06-30
NO-TICKET - refactor and migrate MainContainerFragment.kt on ViewPager2

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/NO-TICKETka

## Description of Work

NO-TICKET - refactor and migrate MainContainerFragment.kt on ViewPager2

### DIFF
--- a/app/src/main/java/org/p2p/wallet/common/ui/BaseFragmentAdapter.kt
+++ b/app/src/main/java/org/p2p/wallet/common/ui/BaseFragmentAdapter.kt
@@ -1,32 +1,34 @@
 package org.p2p.wallet.common.ui
 
-import android.os.Bundle
-import android.util.SparseArray
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentManager
 import androidx.lifecycle.Lifecycle
 import androidx.viewpager2.adapter.FragmentStateAdapter
-import org.p2p.wallet.utils.instantiate
+import android.util.SparseArray
 import kotlin.reflect.KClass
+import org.p2p.wallet.utils.instantiate
 
 class BaseFragmentAdapter(
     private val fragmentManager: FragmentManager,
     lifecycle: Lifecycle,
-    private val items: List<KClass<out Fragment>>,
-    var args: List<Bundle?>? = null
+    private val pages: List<FragmentPageConfiguration>,
 ) : FragmentStateAdapter(fragmentManager, lifecycle) {
 
     val fragments = SparseArray<Fragment>()
 
-    override fun getItemCount(): Int = items.size
+    override fun getItemCount(): Int = pages.size
 
-    override fun createFragment(position: Int): Fragment = instantiateFragment(items[position]).also { fragment ->
-        fragments.put(position, fragment)
-        args?.getOrNull(position)?.let {
-            fragment.arguments = it
+    override fun createFragment(position: Int): Fragment {
+        val pageConfig = pages[position]
+        return instantiateFragment(pageConfig.kClass).also { fragment ->
+            fragments.put(position, fragment)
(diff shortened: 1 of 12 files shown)


eduardmaximovich (approved):

app/src/main/java/org/p2p/wallet/home/ui/container/MainContainerPresenter.kt
```diff
@@ -34,6 +41,34 @@ class MainContainerPresenter(
...
     }
 
+    override fun loadMainNavigation() {
+        val screensConfiguration = getScreenConfiguration()
+        val screensConfigurationMap = screensConfiguration.associate { it.screen to it.kClass }
+        val screensConfigurationArguments = screensConfiguration.map { it.bundle }
+        view?.setMainNavigationConfiguration(screensConfigurationMap, screensConfigurationArguments)
+    }
+
+    private fun getScreenConfiguration() = buildList {
```
eduardmaximovich, line 51: please, add explicit return type

app/src/main/java/org/p2p/wallet/common/ui/BaseFragmentAdapter.kt
```diff
@@ -13,10 +13,10 @@ class BaseFragmentAdapter(
     private val fragmentManager: FragmentManager,
     lifecycle: Lifecycle,
     private val items: List<KClass<out Fragment>>,
-    var args: List<Bundle>? = null
+    var args: List<Bundle?>? = null
```
gslevinkov, line 16: do we really need to make Bundle nullable instead of just empty list?

app/src/main/java/org/p2p/wallet/home/ui/container/MainContainerPresenter.kt
```diff
@@ -34,6 +41,34 @@ class MainContainerPresenter(
         observeRefreshEvent()
     }
 
+    override fun loadMainNavigation() {
+        val screensConfiguration = getScreenConfiguration()
+        val screensConfigurationMap = screensConfiguration.associate { it.screen to it.kClass }
+        val screensConfigurationArguments = screensConfiguration.map { it.bundle }
```
gslevinkov, line 47:
```suggestion
        val screensConfigurationArguments = screensConfiguration.mapNotNull { it.bundle }
```

app/src/main/java/org/p2p/wallet/jupiter/ui/main/JupiterSwapFragment.kt
```diff
@@ -91,6 +90,12 @@ class JupiterSwapFragment :
                     EXTRA_INITIAL_AMOUNT_A to amountA,
                     EXTRA_OPENED_FROM to source
                 )
+
+        fun createBundle(token: Token.Active? = null, source: SwapOpenedFrom): Bundle =
```
gslevinkov, line 94: createArgs

gslevinkov (approved):
looks very nice!!! thanks

app/src/main/java/org/p2p/wallet/common/ui/BaseFragmentAdapter.kt
kabramovp2p: it is because it is related to 2 lists
first one is list of classes second one is list on args

I think we can refactor it and make object which contains class and bundle if needed and it will looks more clear

what do you think ?

app/src/main/java/org/p2p/wallet/home/ui/container/MainContainerPresenter.kt
kabramovp2p: no no we can't do it we need this nulls

PR #1918, opened 2023-07-04
PWN-9158 - hasDeviceShare or same device to check that we can manage device share

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9158

## Description of Work

PWN-9158 - hasDeviceShare or same device to check that we can manage device share

### DIFF
--- a/app/src/main/java/org/p2p/wallet/auth/interactor/MetadataInteractor.kt
+++ b/app/src/main/java/org/p2p/wallet/auth/interactor/MetadataInteractor.kt
@@ -2,7 +2,9 @@ package org.p2p.wallet.auth.interactor
 
 import timber.log.Timber
 import kotlinx.coroutines.CancellationException
+import org.p2p.core.crypto.toBase58Instance
 import org.p2p.solanaj.core.Account
+import org.p2p.solanaj.core.toBase58Instance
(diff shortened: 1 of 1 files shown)


gslevinkov (approved):

1sakov (approved):

app/src/main/java/org/p2p/wallet/auth/interactor/MetadataInteractor.kt
```diff
@@ -60,14 +60,13 @@ class MetadataInteractor(
...
         val metadata = currentMetadata ?: return false
-
         // if device share exist, then we ignore comparing
         if (hasDeviceShare()) {
             return false
         }
 
         // if device share is not empty we are checking with the current system device share
-        return DeviceInfoHelper.getCurrentDeviceName() != metadata.deviceShareDeviceName
+        return !hasDeviceShare() || DeviceInfoHelper.getCurrentDeviceName() != metadata.deviceShareDeviceName
```
1sakov, line 69: wait, `hasDeviceShare` is already checked before

1sakov (changes requested):
i think your code doesn't do anything

1sakov (approved):

PR #1929, opened 2023-07-05
PWN-8901 - Extend Striga cache lifespan until logout

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8901

## Description of Work

- see the comments

image


app/src/main/java/org/p2p/wallet/common/EncryptedSharedPreferences.kt
```diff
@@ -0,0 +1,203 @@
...
+import timber.log.Timber
+import kotlin.properties.ReadWriteProperty
+import kotlin.reflect.KClass
+import kotlin.reflect.KProperty
+import org.p2p.core.crypto.Hex
+import org.p2p.wallet.common.crypto.keystore.DecodeCipher
+import org.p2p.wallet.common.crypto.keystore.EncodeCipher
+import org.p2p.wallet.common.crypto.keystore.KeyStoreWrapper
+
+class EncryptedSharedPreferences(
```
gslevinkov, line 16: instead of randomly using AccountStorage / SecureStorage / smth else we have EncryptedSharedPreferences that has all the logic that encodes / decodes data that comes to shared prefs
no logic inside regarding prefs / namings / cipher is touched.

app/src/main/java/org/p2p/wallet/infrastructure/account/AccountStorage.kt
```diff
@@ -1,73 +1,46 @@
...
-    private val gson: Gson
+    private val encryptedPrefs: EncryptedSharedPreferences
 ) : AccountStorageContract {
 
+    init {
+        encryptedPrefs.loggingTag = "AccountStorage"
+    }
+
     override fun <T> saveObject(key: Key, data: T) {
-        tryWithLog(key) {
```
gslevinkov, line 20: all this logic is the same as in SecureStorage. all extracted to encryptedPrefs

app/src/main/java/org/p2p/wallet/infrastructure/security/SecureStorage.kt
```diff
@@ -1,137 +1,84 @@
...
+    override fun contains(key: Key): Boolean = encryptedPrefs.contains(key.prefsValue)
 
     override fun clear() {
-        sharedPreferences.edit { clear() }
+        encryptedPrefs.clear()
     }
 
     override fun getEncodeCipher(key: Key): EncodeCipher {
-        return tryWithLog(key) { keyStoreWrapper.getEncodeCipher(key.prefsValue) }
+        return encryptedPrefs.getEncodeCipher(key.prefsValue)
```
gslevinkov, line 78: as you can see, no names changed, old users won't be touched

app/src/main/java/org/p2p/wallet/striga/user/StrigaStorage.kt
```diff
@@ -1,44 +1,58 @@
...
+    )
+
+    override var userWallet: StrigaUserWallet? by ObjectEncryptedPreference(
+        preferences = encryptedPrefs,
+        key = KEY_USER_WALLET,
+        type = StrigaUserWallet::class,
+        nullIfMappingFailed = true
+    )
+
+    override var fiatAccount: StrigaFiatAccountDetails? by ObjectEncryptedPreference(
```
gslevinkov, line 40: the final result using new way. 
we explicitly say that we are using encrypted prefs
clear is called on authlogout and this storage has it's own prefs file

app/src/main/java/org/p2p/wallet/common/EncryptedSharedPreferences.kt
gslevinkov: the architecture scheme is on the way

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/common/EncryptedSharedPreferences.kt
```diff
@@ -0,0 +1,255 @@
...
+    private val keyStoreWrapper: KeyStoreWrapper,
+    private val sharedPreferences: SharedPreferences,
+    private val gson: Gson
+) {
+    var loggingTag: String = "EncryptedSharedPreferences"
```
Davran331122, line 21: should be const val

Davran331122: and can be just TAG

app/src/main/java/org/p2p/wallet/common/EncryptedSharedPreferences.kt
```diff
@@ -0,0 +1,255 @@
...
+    fun <T : Any> getObjectList(key: String): List<T> {
+        return tryWithLog(key) {
+            sharedPreferences.getString(key, null)
+                ?.let { keyStoreWrapper.decode(key, it) }
+                ?.let { gson.fromJson(it, object : TypeToken<List<T>>() {}.type) }
+                ?: emptyList()
+        }
+    }
+
+    fun saveBytes(key: String, data: ByteArray) {
```
eduardmaximovich, line 91: why `saveBytes` but `putBoolean`, `putLong` etc? I think the naming should be revised

app/src/main/java/org/p2p/wallet/common/EncryptedSharedPreferences.kt
```diff
@@ -0,0 +1,255 @@
...
+class EncryptedSharedPreferences(
+    private val keyStoreWrapper: KeyStoreWrapper,
+    private val sharedPreferences: SharedPreferences,
+    private val gson: Gson
+) {
+    var loggingTag: String = "EncryptedSharedPreferences"
+
+    fun saveString(key: String, data: String) {
+        tryWithLog(key) {
+            val encodedData = keyStoreWrapper.encode(key, data)
```
Davran331122, line 25: I think, this func should be suspended, and use computation or io dispatcher, cause here we use main thread,
to encode and decode data(and encrypt it)

app/src/main/java/org/p2p/wallet/common/EncryptedSharedPreferences.kt
gslevinkov: sorry, but no
you can write your own tag here for your storage.
i'll write a kdoc

app/src/main/java/org/p2p/wallet/common/EncryptedSharedPreferences.kt
```diff
@@ -0,0 +1,255 @@
...
+            sharedPreferences.edit { putString(key, encodedData) }
+        }
+    }
+
+    fun saveString(key: String, data: String, cipher: EncodeCipher) {
+        tryWithLog(key) {
```
Davran331122, line 31: tryWithLog it's not obvious
you can inject logger to the func, and name it as saveStringAndLog, to avoid using callbacks

app/src/main/java/org/p2p/wallet/common/EncryptedSharedPreferences.kt
gslevinkov: all the Storages and code will be broken if we add some `suspend` stuff here, encoding/decoding is heavy operation, but i didn't see any perfomance issues so far

app/src/main/java/org/p2p/wallet/common/EncryptedSharedPreferences.kt
gslevinkov: sure, let me think of something here

app/src/main/java/org/p2p/wallet/common/EncryptedSharedPreferences.kt
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    fun getString(key: String, cipher: DecodeCipher): String? {
+        return tryWithLog(key) {
+            val encodedData = sharedPreferences.getString(key, null)
+            encodedData?.let { keyStoreWrapper.decode(cipher, it) }
+        }
+    }
+
+    fun <T> saveObject(key: String, data: T) {
```
Davran331122, line 51: this function should be inlined cause generic erase data type after obsuscation, and maybe here you will get an exception, cause type of class was cleared and it also should be suspend

app/src/main/java/org/p2p/wallet/common/EncryptedSharedPreferences.kt
```diff
@@ -0,0 +1,255 @@
...
+            val encodedData = keyStoreWrapper.encode(key, objectAsJson)
+            sharedPreferences.edit { putString(key, encodedData) }
+        }
+    }
+
+    /**
+     * !!! WARNING !!!
+     * DOESN'T WORK WITH CLASSES
+     */
+    @Deprecated(message = "DOESN'T WORK, use only with primitive types like String")
```
Davran331122, line 63: If this class works only with primitives,
you can create something  like 
`@JvmInline
value class NonZeroInt(val value: Int) {
    init {
        require(value != 0) { "0 is not a value for this type!" }
    }
}
`

Davran331122: I think annotate new method with @Deprecated annotation is not good idea

app/src/main/java/org/p2p/wallet/common/EncryptedSharedPreferences.kt
```diff
@@ -0,0 +1,255 @@
...
+    fun <T> saveObjectList(key: String, data: List<T>) {
+        saveObject(key, data)
+    }
+
+    @Throws(IllegalArgumentException::class)
+    fun <T : Any> getObject(key: String, type: KClass<T>): T? {
```
Davran331122, line 69: and this function also should be suspend, and I think we remove here @Throw annotation and  handle it's here to return just null

app/src/main/java/org/p2p/wallet/common/EncryptedSharedPreferences.kt
```diff
@@ -0,0 +1,255 @@
...
+        return tryWithLog(key) {
+            sharedPreferences.getLong(key, defaultValue)
+        }
+    }
+
+    /**
+     * Encodes each element inside the set and saves it to the shared preferences.
+     * Set is unique by definition, so when elements are encoded - they are unique too.
+     */
+    fun putStringSet(key: String, value: Set<String>) {
```
Davran331122, line 137: I think, we can use List here, cause preferences does not work with set, and Kotlin.Set is MutableCollection
so I don't see reasons to have this method

app/src/main/java/org/p2p/wallet/home/ui/main/HomeInteractor.kt
```diff
@@ -44,7 +44,7 @@ class HomeInteractor(
         if (strigaSignupEnabledFeatureToggle.isFeatureEnabled) {
             strigaSignupInteractor.loadAndSaveSignupData()
             strigaUserInteractor.loadAndSaveUserStatusData()
-            if (strigaUserInteractor.isUserCreated()) {
+            if (strigaUserInteractor.isUserCreated() && strigaUserInteractor.isKycApproved) {
```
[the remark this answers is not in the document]
gslevinkov: already done in other pr

app/src/main/java/org/p2p/wallet/common/EncryptedSharedPreferences.kt
gslevinkov: string set is supported by preferences, you can check android.content.SharedPreferences#getStringSet

app/src/main/java/org/p2p/wallet/striga/user/StrigaStorage.kt
```diff
@@ -1,61 +1,80 @@
...
+        keyProvider = { KEY_USER_STATUS },
+        type = StrigaUserStatusDetails::class,
+        nullIfMappingFailed = true
+    )
+
+    override var userWallet: StrigaUserWallet? by ObjectEncryptedPreference(
+        preferences = encryptedPrefs,
+        keyProvider = { KEY_USER_WALLET },
+        type = StrigaUserWallet::class,
+        nullIfMappingFailed = true
```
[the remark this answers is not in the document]
gslevinkov: we don't want to crash or show the error if this mapping if failed
so if it's null on the caller side - we shall fetch the data again and save it

EncryptedStoragePrefs can throw some exceptions - on mapping for example or working with KeyStore

app/src/main/java/org/p2p/wallet/striga/user/StrigaStorage.kt
```diff
@@ -1,61 +1,80 @@
...
+    )
+
+    override var smsExceededVerificationAttemptsMillis: MillisSinceEpoch by LongPreference(
+        preferences = encryptedPrefs,
+        keyProvider = { KEY_SMS_EXCEEDED_VERIFICATION_ATTEMPTS_MILLIS },
+        defaultValue = 0
+    )
+
+    override var smsExceededResendAttemptsMillis: MillisSinceEpoch by LongPreference(
```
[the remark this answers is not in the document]
gslevinkov: 0 was used by Eduard before, so I just used it here too

app/src/main/java/org/p2p/wallet/striga/user/StrigaStorage.kt
```diff
@@ -1,61 +1,80 @@
...
-        sharedPreferences.edit(true) {
-            putBoolean(KEY_USER_BANNER_IS_HIDDEN_PREFIX + banner.name, true)
-        }
+        val hiddenBanners = encryptedPrefs.getStringSet(KEY_USER_BANNER_IS_HIDDEN)
+        encryptedPrefs.putStringSet(KEY_USER_BANNER_IS_HIDDEN, hiddenBanners + banner.bannerId.toString())
     }
 
     override fun isBannerHidden(banner: StrigaKycStatusBanner): Boolean {
-        return sharedPreferences.getBoolean(KEY_USER_BANNER_IS_HIDDEN_PREFIX + banner.name, false)
+        return banner.toString() in encryptedPrefs.getStringSet(KEY_USER_BANNER_IS_HIDDEN)
```
Davran331122, line 68: can you separate this in two lines ? to have ability to take it easy on debugging ?

gslevinkov: yep, will do

app/src/main/java/org/p2p/wallet/common/EncryptedSharedPreferences.kt
eduardmaximovich: I think you right regarding exceptions, this function should return null on error because all these errors cannot be recovered and should be just logged and then fixed.

Suspend function is an overhead for now, it would require to refactor essential part of app and re-check everything

eduardmaximovich (approved):

PR #1933, opened 2023-07-06
PWN-9202 - striga. fixed navigation to pending bottomsheet

eduardmaximovich:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9202

## Description of Work

- Fixed navigation to kyc pedning when clicked bank transfer

### DIFF
--- a/app/src/main/java/org/p2p/wallet/striga/user/repository/StrigaUserStatusDestinationMapper.kt
+++ b/app/src/main/java/org/p2p/wallet/striga/user/repository/StrigaUserStatusDestinationMapper.kt
@@ -17,8 +17,11 @@ class StrigaUserStatusDestinationMapper {
             userStatus?.isMobileVerified == false -> {
                 StrigaUserStatusDestination.SMS_VERIFICATION
             }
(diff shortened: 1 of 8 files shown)


kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/striga/user/repository/StrigaUserStatusDestinationMapper.kt
```diff
@@ -17,9 +17,12 @@ class StrigaUserStatusDestinationMapper {
             userStatus?.isMobileVerified == false -> {
```
Davran331122, line 17: you can create local not nullable variable 
`val status = userStatus ?: return `
and avoid compare nullable value with true or false

eduardmaximovich: I'm going to remove isUserNotCreated and use userStatus directly to eliminate nullable comparisons through automatic non-null casting

PR #1938, opened 2023-07-06
PWN-9204 - striga. update banner images

eduardmaximovich:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9204

## Description of Work

- Updated kyc banner images


1sakov (approved):

app/src/main/res/drawable/ic_striga_kyc_approved.xml
```diff
@@ -0,0 +1,101 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
gslevinkov, line 1: you sure it's a good idea to keep this big PNG as SVG?

gslevinkov (approved):

app/src/main/res/drawable/ic_striga_kyc_approved.xml
eduardmaximovich: @gslevinkov I've checked, it doesn't have any artifacts etc, so why not? some of such big images we store in vector, and we don't face any problems with them

PR #1952, opened 2023-07-12
NO-TICKET - Minor improvements for token services

1sakov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/

## Description of Work

## Screenshots (Optional)


app/src/main/java/org/p2p/wallet/home/ui/wallet/WalletPresenter.kt
```diff
@@ -24,58 +27,69 @@ class WalletPresenter(
...
+        view?.showRefreshing(isRefreshing = newState.isLoading())
+
+        when (newState) {
+            is UserTokensState.Idle -> Unit
+            is UserTokensState.Loading -> Unit
+            is UserTokensState.Refreshing -> Unit
+            is UserTokensState.Error -> view?.showErrorMessage(newState.cause)
+            is UserTokensState.Empty -> view?.showBalance(null)
+            is UserTokensState.Loaded -> {
+                val usdc = newState.solTokens.find { it.isUSDC }
```
kabramovp2p, line 65: but we should get all tokens with needed extras not only USDC :(

gslevinkov, line 65: isUsdc

kabramovp2p (approved):

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/home/ui/wallet/WalletPresenter.kt
1sakov: For crypto, right?
For Wallet you need only usdc

1sakov: why? 
The result of `find` is token, not boolean

gslevinkov: { it.isUSDC } -> { it.isUsdc } :)

PR #1955, opened 2023-07-12
PWN-9151 - MyCryptoFragment.kt created and logic added

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9151

## Description of Work

PWN-9151 - MyCryptoFragment.kt created and logic added


app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoFragment.kt
```diff
@@ -0,0 +1,174 @@
...
+            }
+            ActionButton.SWAP_BUTTON -> {
+                presenter.onSwapClicked()
+            }
+            else -> Timber.d("Unsupported Action! $clickedButton")
+        }
+    }
+
+    override fun showBalance(cellModel: TextViewCellModel?) {
+        binding.viewBalance.textViewAmount.bindOrGone(cellModel)
```
gslevinkov, line 120: balance can be empty if there is no usdc?

app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoFragment.kt
```diff
@@ -0,0 +1,174 @@
...
+
+    override fun showReceive() {
+        replaceFragment(receiveFragmentFactory.receiveFragment())
+    }
+
+    override fun showSwap() {
+        replaceFragment(JupiterSwapFragment.create(source = SwapOpenedFrom.MAIN_SCREEN))
+    }
+
+    private fun onToggleClicked() {
```
gslevinkov, line 148: onVisibilityToggleClicked

on the first view i thought about feature toggle 😁

app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoFragment.kt
```diff
@@ -0,0 +1,174 @@
...
+    private fun onHideClicked(token: Token.Active) {
+        homeAnalytics.logHiddenTokensClicked()
+        presenter.toggleTokenVisibility(token)
+    }
+
+    private fun onClaimTokenClicked(canBeClaimed: Boolean, token: Token.Eth) {
+        presenter.onClaimClicked(canBeClaimed, token)
+    }
+
+    override fun showTokenClaim(token: Token.Eth) {
```
gslevinkov, line 167: navigateToTokenClaim

app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoPresenter.kt
```diff
@@ -0,0 +1,195 @@
...
+            visibilityState = visibilityState,
+            isZerosHidden = areZerosHidden,
+        )
+        view?.showItems(mappedItems)
+    }
+
+    private fun getUserBalance(tokens: List<Token.Active>): BigDecimal {
+        if (tokens.none { it.totalInUsd != null }) return BigDecimal.ZERO
+
+        return tokens
```
gslevinkov, line 108: UserRepository already has this logic

app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoPresenter.kt
```diff
@@ -0,0 +1,195 @@
...
+                    solTokens = newState.solTokens,
+                    ethTokens = newState.ethTokens
+                )
+            }
+        }
+    }
+
+    private fun showTokensAndBalance(solTokens: List<Token.Active>, ethTokens: List<Token.Eth>) {
+        val balance = getUserBalance(solTokens)
+        view?.showBalance(homeMapper.mapBalance(balance))
```
gslevinkov, line 92: showBalance can get only non-null balance

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoPresenter.kt
```diff
@@ -0,0 +1,195 @@
...
+            is UserTokensState.Empty -> {
+                view?.showEmptyState(isEmpty = true)
+                handleEmptyAccount()
+            }
+            is UserTokensState.Loaded -> {
+                view?.showEmptyState(isEmpty = false)
+                initializeActionButtons()
+                showTokensAndBalance(
+                    solTokens = newState.solTokens,
+                    ethTokens = newState.ethTokens
```
1sakov, line 84: Let's get rid of old functions like `initializeActionButtons()` or `showTokensAndBalance`

1sakov: Can we try to make a single `CryptoState` which can modify the UI?
I want only one channel which edits the UI

1sakov (changes requested):

app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoPresenter.kt
kabramovp2p: yes that is true

app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoPresenter.kt
kabramovp2p: yes but we will need to filter this tokens on USDC and other Wallet tokens in future

1sakov (approved):

PR #1963, opened 2023-07-13
NO-TICKET - Revert metadata saving to JSON

Davran331122:


app/src/main/java/org/p2p/wallet/common/storage/ExternalStorageRepository.kt
```diff
@@ -1,17 +1,19 @@
...
-    fun saveJson(json: String, fileName: String) {
+
+    fun <T> saveJson(jsonObject: T, fileName: String){
         try {
-            context.openFileOutput(fileName, Context.MODE_PRIVATE).use {
-                it.write(json.toByteArray())
-            }
+            context.openFileOutput(fileName, Context.MODE_PRIVATE)
+                .bufferedWriter()
+                .use { Gson().toJson(jsonObject, it) }
```
gslevinkov, line 16: provide Gson to constructor

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/user/interactor/TokenMetadataInteractor.kt
```diff
@@ -10,7 +10,7 @@ import org.p2p.wallet.common.storage.ExternalStorageRepository
 import org.p2p.wallet.user.repository.UserLocalRepository
 
 private const val TAG = "TokenMetadataInteractor"
-private const val TOKENS_FILE_NAME = "tokens"
+private const val TOKENS_FILE_NAME = "tokens_metadata"
```
eduardmaximovich, line 13: I assume that versioning could be a good idea, otherwise if something changes we will need to come up with a new value. What do you think?
```suggestion
private const val TOKENS_FILE_NAME = "tokens_metadata_v1"
```

eduardmaximovich (approved):

app/src/main/java/org/p2p/wallet/user/interactor/TokenMetadataInteractor.kt
Davran331122: I think there is no need to do thiы

PR #1967, opened 2023-07-14
PWN-8987 - PWN-9277 - Metadata last modified fixed

1sakov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-8987
https://p2pvalidator.atlassian.net/browse/PWN-9277

## Description of Work

## Screenshots (Optional)

### DIFF
--- a/app/src/main/java/org/p2p/wallet/user/interactor/TokenMetadataInteractor.kt
+++ b/app/src/main/java/org/p2p/wallet/user/interactor/TokenMetadataInteractor.kt
@@ -10,7 +10,7 @@ import org.p2p.wallet.common.storage.ExternalStorageRepository
 import org.p2p.wallet.user.repository.UserLocalRepository
 
 private const val TAG = "TokenMetadataInteractor"
-private const val TOKENS_FILE_NAME = "tokens_metadata"
+const val TOKENS_FILE_NAME = "tokens_metadata"
 
 class TokenMetadataInteractor(
     private val externalStorageRepository: ExternalStorageRepository,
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/p2p/wallet/user/interactor/TokenMetadataInteractor.kt
```diff
@@ -27,20 +27,22 @@ class TokenMetadataInteractor(
             null
         }
 
-        val lastModified = metadata?.timestamp
-        Timber.tag(TAG).i("Checking if metadata is modified since: $lastModified")
+        val ifModifiedSince = metadata?.timestamp
```
gslevinkov, line 30: without if please :)

token-service/src/main/java/org/p2p/token/service/repository/metadata/TokenMetadataRemoteRepository.kt
```diff
@@ -25,29 +25,31 @@ internal class TokenMetadataRemoteRepository(
...
-                } else {
-                    UpdateTokenMetadataResult.Error(e)
-                }
+    override suspend fun loadTokensMetadata(
+        ifModifiedSince: String?
+    ): UpdateTokenMetadataResult = withContext(dispatchers.io) {
+        val response = try {
+            api.getZipFile("${tokenServiceUrl}get_all_tokens_info", ifModifiedSince)
+        } catch (e: HttpException) {
+            return@withContext if (e.code() == HttpURLConnection.HTTP_NOT_MODIFIED) {
```
gslevinkov, line 34: i think return@withContext is not really needed here

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/user/interactor/TokenMetadataInteractor.kt
1sakov: it's the header name. 
`if-modified-since`. Why should I remove it?

token-service/src/main/java/org/p2p/token/service/repository/metadata/TokenMetadataRemoteRepository.kt
1sakov: why? 
We should return the function. 

The code below expects the response only

kabramovp2p (approved):

Davran331122 (approved):

PR #1970, opened 2023-07-14
PWN-9151 - HomeFragment removed

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9151

## Description of Work

PWN-9151 - HomeFragment removed


app/src/main/java/org/p2p/wallet/user/interactor/UserInteractor.kt
```diff
@@ -19,6 +18,7 @@ import org.p2p.wallet.user.repository.UserTokensLocalRepository
 import org.p2p.wallet.utils.emptyString
 
 private const val KEY_HIDDEN_TOKENS_VISIBILITY = "KEY_HIDDEN_TOKENS_VISIBILITY"
+val TOKEN_SYMBOLS_VALID_FOR_BUY: List<String> = listOf(Constants.USDC_SYMBOL, Constants.SOL_SYMBOL)
```
gslevinkov, line 21: maybe Set<String> would be better here?

gslevinkov: and it's better to use mint addresses instead of symbols
symbols can change

gslevinkov (approved):

1sakov (approved):

PR #1971, opened 2023-07-15
NO-TICKET - FeatureToggleProviderTest added

gslevinkov:


core/src/main/java/org/p2p/core/network/gson/GsonTypeAdapters.kt
```diff
@@ -184,7 +184,7 @@ internal class DefaultBlockParameterTypeAdapter : TypeAdapter<DefaultBlockParame
 
 
 
-internal class OptionalTypeAdapter<T>(
+internal class OptionalTypeAdapter<T : Any>(
```
gslevinkov, line 187: remove kotlin compiler warning

1sakov (approved):

kabramovp2p:
thanks a lot for the tests!

kabramovp2p (approved):

PR #1973, opened 2023-07-17
NO-TICKET - Run ktlint to all modules in app

gslevinkov:
only line breaks and spaces, nothing more.

### DIFF
--- a/core/src/main/java/org/p2p/core/wrapper/eth/rlp/RLP.kt
+++ b/core/src/main/java/org/p2p/core/wrapper/eth/rlp/RLP.kt
@@ -5,7 +5,7 @@ import org.bouncycastle.util.BigIntegers.asUnsignedByteArray
 import org.bouncycastle.util.encoders.Hex
 import java.io.Serializable
 import java.math.BigInteger
-import java.util.*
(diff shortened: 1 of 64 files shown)


core/src/main/java/org/p2p/core/wrapper/eth/rlp/RLP.kt
```diff
@@ -303,15 +329,22 @@ object RLP {
...
+                        /* length = */ lengthOfLength.toInt() + length + 1
+                    )
 
                     if (level + 1 < depth) {
                         val newLevelList = RLPList()
                         newLevelList.rlpData = rlpData
 
-                        fullTraverse(msgData, level + 1, pos + lengthOfLength.toInt() + 1,
-                                pos + lengthOfLength.toInt() + length + 1, newLevelList, depth)
+                        fullTraverse(
```
gslevinkov, line 344: if you are scared that ETH module will stop working - i can add `internal` directory for ETH and add an exclude rule for ktlint

1sakov (approved):

kabramovp2p (approved):

PR #1974, opened 2023-07-17
PWN-9281 - kover multimodule reports - final version

eduardmaximovich:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9281

## Description of Work

Multimodule reports for kover.
Included only 2 extra modules (core + token-service) as we are not ready to test other modules: ethereumkit, solana, ui-kit


1sakov (approved):

core/build.gradle.kts
```diff
@@ -1,6 +1,34 @@
 plugins {
     id("org.p2p.wallet.android.library")
     id("kotlin-kapt")
+    id("org.jetbrains.kotlinx.kover") version "0.7.2"
+}
+
+koverReport {
```
eduardmaximovich, line 7: it doesn't make sense to move these configs to buildSrc as every module has its own exclusion config etc

gslevinkov (approved):

PR #1979, opened 2023-07-18
PWN-9197 - Add metadata ext, and add saving ext in database(Part 1)

Davran331122:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9197


core/src/main/java/org/p2p/core/token/Token.kt
```diff
@@ -42,14 +42,15 @@ sealed class Token constructor(
         val totalInUsd: BigDecimal?,
         val total: BigDecimal,
         val visibility: TokenVisibility,
+        val extensions: TokenExtension?,
```
gslevinkov, line 45: TokenExtension -> TokenExtensions

app/src/main/java/org/p2p/wallet/user/local/TokenListResponse.kt
```diff
@@ -1,42 +0,0 @@
-package org.p2p.wallet.user.local
-
-import com.google.gson.annotations.SerializedName
-
-data class TokenListResponse(
-
-    @SerializedName("tokens")
-    val tokens: List<TokenResponse>
```
1sakov, line 8: Why this class is deleted? 
It's being used while getting token metadata

ethereumkit/src/main/java/org/p2p/ethereumkit/external/repository/EthereumKitRepository.kt
```diff
@@ -99,11 +98,14 @@ internal class EthereumKitRepository(
                 )
             }
                 .filter { token ->
-                    val tokenBundle = claimingTokens.firstOrNull { token.publicKey == it.contractAddress.hex }
-                    val tokenFiatAmount = token.totalInUsd.orZero()
-                    val isClaimInProgress = tokenBundle != null && tokenBundle.isClaiming
-                    tokenFiatAmount >= MINIMAL_DUST || isClaimInProgress
+                    true
+//                    val tokenBundle = claimingTokens.firstOrNull { token.publicKey == it.contractAddress.hex }
```
gslevinkov: oops?

app/src/main/java/org/p2p/wallet/user/local/TokenListResponse.kt
Davran331122: yeah, we have copy of this class in token-service module

gslevinkov (approved):

PR #1981, opened 2023-07-18
PWN-9265 - Striga. /account/statement

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9265

## Description of Work

- split one big wallet repository into two different repo
- implement api request with all needed info

### DIFF
--- a/app/src/main/java/org/p2p/wallet/striga/wallet/repository/impl/StrigaWalletCachedRepository.kt
+++ b/app/src/main/java/org/p2p/wallet/striga/wallet/repository/impl/StrigaWalletCachedRepository.kt
@@ -60,7 +60,7 @@ internal class StrigaWalletCachedRepository(
     ): StrigaDataLayerResult<T> {
         return try {
             val result = cache?.invoke()?.also {
-                timber.i("Getting from cache getting data")
+                timber.i("Getting data from cache")
             } ?: remote().also {
                 timber.i("Update cache from remote")
                 cache?.set(it)
--- a/app/src/main/java/org/p2p/wallet/striga/wallet/repository/impl/StrigaWalletCachedRepository.kt
+++ b/app/src/main/java/org/p2p/wallet/striga/wallet/repository/impl/StrigaWalletCachedRepository.kt
@@ -1,0 +1,76 @@
+package org.p2p.wallet.striga.wallet.repository.impl
+
+import timber.log.Timber
+import kotlin.reflect.KMutableProperty0
+import org.p2p.wallet.striga.common.model.StrigaDataLayerError
+import org.p2p.wallet.striga.common.model.StrigaDataLayerResult
+import org.p2p.wallet.striga.common.model.toSuccessResult
+import org.p2p.wallet.striga.user.repository.StrigaUserRepository
+import org.p2p.wallet.striga.wallet.models.StrigaCryptoAccountDetails
+import org.p2p.wallet.striga.wallet.models.StrigaFiatAccountDetails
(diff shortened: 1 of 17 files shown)


app/src/main/java/org/p2p/wallet/striga/wallet/repository/impl/StrigaWalletCachedRepository.kt
```diff
@@ -0,0 +1,76 @@
...
+    private val remoteRepository: StrigaWalletRemoteRepository,
+    private val cache: StrigaWalletInMemoryRepository,
+    private val strigaUserRepository: StrigaUserRepository,
+) : StrigaWalletRepository {
+
+    private val timber: Timber.Tree = Timber.tag("StrigaWalletCachedRepository")
+
+    override suspend fun getUserWallet(): StrigaDataLayerResult<StrigaUserWallet> {
+        return withCache(cache::userWallet) {
+            remoteRepository.getUserWallet()
```
gslevinkov, line 26: look how clean this is!

app/src/main/java/org/p2p/wallet/striga/wallet/repository/impl/StrigaWalletCachedRepository.kt
```diff
@@ -0,0 +1,76 @@
...
+    ): StrigaDataLayerResult<StrigaCryptoAccountDetails> {
+        return withCache(cache::cryptoAccountDetails) {
+            remoteRepository.getCryptoAccountDetails(accountId)
+        }
+    }
+
+    override suspend fun getAccountStatement(
+        accountId: StrigaAccountId,
+    ): StrigaDataLayerResult<StrigaUserBankingDetails> {
+        return withCache(null) {
```
gslevinkov, line 49: no need to cache now

kabramovp2p: i mean why not `withCache() {` ?

eduardmaximovich: @kabramovp2p I think the argument should be set explicitly, because you might think that if you're not passing any argument, it will use some default variable

gslevinkov: yeah, absolutely 
you can see explicitly that this method uses no cache - so null is passed. 
```kotlin
withCache {
    ...
}
```
looks that there cache anyway

eduardmaximovich: @gslevinkov you know, I've just realized, why is there withCache called if the cache is not used at all? maybe this method should just return the result directly?
```
override suspend fun getAccountStatement(
     accountId: StrigaAccountId,
): StrigaDataLayerResult<StrigaUserBankingDetails> {
    val strigaUserFullName = strigaUserRepository.getUserDetails().unwrap()
        .userInfo
        .fullName
    return remoteRepository.getAccountStatement(accountId, strigaUserFullName)
}
```

eduardmaximovich (approved):

app/src/main/java/org/p2p/wallet/striga/wallet/repository/impl/StrigaWalletCachedRepository.kt
gslevinkov: yeah, but we have unwrap here that will break the contract

PR #1982, opened 2023-07-19
PWN-9286 - NFT got decimals == 0 so we can filter them

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9286

## Description of Work

PWN-9286 - NFT got decimals == 0 so we can filter them


1sakov (approved):

app/src/main/java/org/p2p/wallet/user/repository/UserTokensRemoteRepository.kt
```diff
@@ -37,6 +36,7 @@ class UserTokensRemoteRepository(
             }
 
             val token = userLocalRepository.findTokenData(mintAddress) ?: return@mapNotNull null
+            if (token.decimals == 0) return@mapNotNull null
```
gslevinkov, line 39: 0 to const

gslevinkov (approved):

PR #1993, opened 2023-07-21
PWN-9305 - make changes according to `add money` logic

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9305

## Description of Work

PWN-9305 - make changes according to add money logic

### DIFF
--- a/app/src/main/java/org/p2p/wallet/home/ui/topup/TopUpWalletPresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/home/ui/topup/TopUpWalletPresenter.kt
@@ -61,6 +61,7 @@ class TopUpWalletPresenter(
                     view?.navigateToBuyWithTransfer(it)
                 }
             }
+            return
         }
         // in case of simulation web3 user, we don't need to check metadata
         if (inAppFeatureFlags.strigaSimulateWeb3Flag.featureValue) {
--- a/app/src/main/java/org/p2p/wallet/home/ui/topup/TopUpWalletPresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/home/ui/topup/TopUpWalletPresenter.kt
@@ -42,7 +42,7 @@ class TopUpWalletPresenter(
             strigaBankTransferProgress.onEach(view::showStrigaBankTransferView)
                 .launchIn(this)
         } else {
-            view.hideStrigaBankTransferView()
+            view.showStrigaBankTransferView(showProgress = false)
         }
 
         launch {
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/p2p/wallet/home/ui/topup/TopUpWalletPresenter.kt
```diff
@@ -54,6 +54,14 @@ class TopUpWalletPresenter(
...
 
     override fun onBankTransferClicked() {
+        if (!strigaSignupFeatureToggle.isFeatureEnabled) {
+            launch {
+                val tokenToBuy = userInteractor.getSingleTokenForBuy()
+                tokenToBuy?.let {
+                    view?.navigateToBuyWithTransfer(it)
+                }
+            }
+        }
```
eduardmaximovich, line 64: I think there is missing `return` statement

gslevinkov: Shouldn't we hide the transfer item if striga is not enabled?

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/moonpay/ui/new/NewBuyPresenter.kt
```diff
@@ -103,7 +104,9 @@ class NewBuyPresenter(
             view?.showLoading(isLoading = true)
             currentAlphaCode = paymentMethodsInteractor.getBankTransferAlphaCode()
 
-            val availablePaymentMethods = paymentMethodsInteractor.getAvailablePaymentMethods(currentAlphaCode)
+            val availablePaymentMethods = paymentMethodsInteractor.getAvailablePaymentMethods(
+                currentAlphaCode, preselectedMethodType
```
eduardmaximovich, line 108: to make it clear, let `alphaCode` variables always represent the standard of their value:
```suggestion
                currentAlpha3Code, preselectedMethodType
```

eduardmaximovich (approved):

app/src/main/java/org/p2p/wallet/home/ui/topup/TopUpWalletPresenter.kt
kabramovp2p: no
in new logic we should not

eduardmaximovich: @gslevinkov hah, yeah, it hides the item in attachView if striga is disabled and user is web3-authorized
@kabramovp2p have you checked this?

PR #2016, opened 2023-07-26
PWN-9362 - fix layout to make needed scroll behavior

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9362

## Description of Work

PWN-9362 - fix layout to make needed scroll behavior

### DIFF
--- a/app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoFragment.kt
+++ b/app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoFragment.kt
@@ -165,16 +165,16 @@ class MyCryptoFragment :
         setAppBarScrollingState(!isEmpty)
     }
 
-    private fun setAppBarScrollingState(isEnabled: Boolean) {
+    private fun setAppBarScrollingState(isScrollingEnabled: Boolean) {
         with(binding) {
             collapsingToolbar.updateLayoutParams<AppBarLayout.LayoutParams> {
-                scrollFlags = if (isEnabled) {
+                scrollFlags = if (isScrollingEnabled) {
                     SCROLL_FLAG_SCROLL
                 } else {
                     SCROLL_FLAG_NO_SCROLL
                 }
             }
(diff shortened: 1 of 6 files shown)


eduardmaximovich (approved):

app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoFragment.kt
```diff
@@ -158,6 +162,22 @@ class MyCryptoFragment :
             textViewEmpty.isVisible = isEmpty
             recyclerViewCrypto.isVisible = !isEmpty
         }
+        setAppBarScrollingState(!isEmpty)
+    }
+
+    private fun setAppBarScrollingState(isEnabled: Boolean) {
```
gslevinkov, line 168: isScrollingEnabled

app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoFragment.kt
```diff
@@ -158,6 +162,22 @@ class MyCryptoFragment :
...
+        with(binding) {
+            collapsingToolbar.updateLayoutParams<AppBarLayout.LayoutParams> {
+                scrollFlags = if (isEnabled) {
+                    SCROLL_FLAG_SCROLL
+                } else {
+                    SCROLL_FLAG_NO_SCROLL
+                }
+            }
+            if (!isEnabled) {
+                appBarLayout.setExpanded(true, false)
```
gslevinkov, line 178: true false😁

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoFragment.kt
kabramovp2p: nothing we can do here =)
it is java method =)))

PR #2019, opened 2023-07-26
PWN-9408 - do not count balance that are less then 0.01

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9408

## Description of Work

PWN-9408 - do not count balance that are less then 0.01


app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoPresenter.kt
```diff
@@ -126,6 +129,7 @@ class MyCryptoPresenter(
 
         return tokens
             .mapNotNull(Token.Active::totalInUsd)
+            .filter { it.isMoreThan(MINIMAL_DUST_FOR_BALANCE) }
```
gslevinkov, line 132: looks like domain logic
do you want maybe extract this getUserBalance to the interactor?

gslevinkov (approved):

PR #2022, opened 2023-07-26
PWN-9352 - Striga. Calculator layout fixes

eduardmaximovich:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9352

## Description of Work

Fixed UI for striga off-ramp calculator

### DIFF
--- a/app/src/main/java/org/p2p/wallet/striga/offramp/mappers/StrigaOffRampSwapWidgetMapper.kt
+++ b/app/src/main/java/org/p2p/wallet/striga/offramp/mappers/StrigaOffRampSwapWidgetMapper.kt
@@ -1,6 +1,7 @@
 package org.p2p.wallet.striga.offramp.mappers
 
 import androidx.annotation.StyleRes
+import android.util.TypedValue
 import java.math.BigDecimal
 import org.p2p.core.common.TextContainer
 import org.p2p.core.utils.STRIGA_FIAT_DECIMALS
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/p2p/wallet/striga/offramp/mappers/StrigaOffRampSwapWidgetMapper.kt
```diff
@@ -75,7 +77,10 @@ class StrigaOffRampSwapWidgetMapper {
             isStatic = isStatic,
             widgetTitle = mapWidgetTitle(titleResId),
             availableAmount = mapAvailableAmount(isLoadingBalance, balance, currencyName),
-            currencyName = currencyName.toRawTextViewCellModel(R.style.UiKit_TextAppearance_Regular_Label1),
+            currencyName = currencyName.toRawTextViewCellModel(
+                textSizeSp = 28f,
```
gslevinkov, line 81: wow
Title1 should have text size already

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/striga/offramp/mappers/StrigaOffRampSwapWidgetMapper.kt
eduardmaximovich: @gslevinkov it has, but nope, it just doesn't work despite SwapWidget [binds the model fully](https://github.com/p2p-org/key-app-android/blob/develop/ui-kit/src/main/kotlin/org/p2p/uikit/utils/text/TextViewCellModel.kt#L122)  and textViewSize by default is null, it uses some other text size, maybe directly from layout
image

Since we have textViewSize field, I honestly assumed that we can override this value directly in the model

kabramovp2p (approved):

PR #2025, opened 2023-07-26
PWN-9232 - app loaders with dependencies

eduardmaximovich:
https://p2pvalidator.atlassian.net/browse/PWN-9232

## Description of Work

- Make AppLoader aware dependencies
- Await only for dependant loaders, otherwise it works in non-blocking mode

### DIFF
--- a/app/src/main/java/org/p2p/wallet/home/events/AppLoader.kt
+++ b/app/src/main/java/org/p2p/wallet/home/events/AppLoader.kt
@@ -1,12 +1,12 @@
 package org.p2p.wallet.home.events
 
 abstract class AppLoader {
-    private val dependencies: MutableList<AppLoader> = mutableListOf()
+    private val dependencies: MutableSet<AppLoader> = mutableSetOf()
     abstract suspend fun onLoad()
     open suspend fun onRefresh(): Unit = Unit
     open suspend fun isEnabled(): Boolean = true
 
-    fun getDependencies(): List<AppLoader> = dependencies
+    fun getDependencies(): Set<AppLoader> = dependencies
 
     fun dependsOn(vararg loaders: AppLoader) {
         dependencies += loaders
--- a/app/src/main/java/org/p2p/wallet/home/events/AppLoader.kt
+++ b/app/src/main/java/org/p2p/wallet/home/events/AppLoader.kt
@@ -1,7 +1,14 @@
 package org.p2p.wallet.home.events
 
-interface AppLoader {
-    suspend fun onLoad()
-    suspend fun onRefresh(): Unit = Unit
-    suspend fun isEnabled(): Boolean = true
+abstract class AppLoader {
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/p2p/wallet/home/events/AppLoader.kt
```diff
@@ -1,7 +1,14 @@
...
 
-interface AppLoader {
-    suspend fun onLoad()
-    suspend fun onRefresh(): Unit = Unit
-    suspend fun isEnabled(): Boolean = true
+abstract class AppLoader {
+    private val dependencies: MutableList<AppLoader> = mutableListOf()
+    abstract suspend fun onLoad()
+    open suspend fun onRefresh(): Unit = Unit
+    open suspend fun isEnabled(): Boolean = true
```
gslevinkov, line 7: I would rather make it as abstract to make loaders more explicit - you clearly see conditions when they are loading and you won't forget to use FT

app/src/main/java/org/p2p/wallet/home/events/AppLoader.kt
```diff
@@ -1,7 +1,14 @@
 package org.p2p.wallet.home.events
 
-interface AppLoader {
-    suspend fun onLoad()
-    suspend fun onRefresh(): Unit = Unit
-    suspend fun isEnabled(): Boolean = true
+abstract class AppLoader {
+    private val dependencies: MutableList<AppLoader> = mutableListOf()
```
gslevinkov, line 4: Maybe Set? To remove duplicate requests

app/src/main/java/org/p2p/wallet/home/events/AppLoader.kt
gslevinkov: Upd: I see, they are unique in the Facade

Davran331122 (approved):

app/src/main/java/org/p2p/wallet/home/events/AppLoaderFacade.kt
```diff
@@ -13,17 +13,70 @@ import org.p2p.core.common.di.AppScope
...
+            val dependencies = loader.getDependencies().mapNotNull { dependency ->
+                appLoaders.find { it.id() == dependency.id() }
+            }
+            dependencyGraph[loader] = dependencies
+        }
+
+        return dependencyGraph
+    }
+
+    private fun AppLoader.id(): String = javaClass.name
```
gslevinkov, line 81: careful with proguard

gslevinkov (approved):

kabramovp2p (approved):

app/src/main/java/org/p2p/wallet/home/events/AppLoaderFacade.kt
eduardmaximovich: I've checked feature build, it seems everything works correctly

PR #2054, opened 2023-07-31
PWN-9293 - striga. api urls fixes

eduardmaximovich:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9293

## Description of Work

Fixed striga urls

### DIFF
--- a/core/src/main/res/values/environment_config.xml
+++ b/core/src/main/res/values/environment_config.xml
@@ -41,7 +41,6 @@
     <string name="alarmErrorsServiceBaseUrl">https://oncall.keyapp.org/</string>
 
     <string name="strigaProxyServiceBaseUrl">https://payment-rust.keyapp.org/</string>
-    <string name="strigaProxyServiceProdBaseUrl">https://payment-rust.key.app/</string>
 
     <string name="tokenServiceBaseUrl">https://token-service.keyapp.org/</string>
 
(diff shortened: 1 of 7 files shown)


core/src/main/res/values/environment_config.xml
```diff
@@ -40,8 +40,8 @@
 
     <string name="alarmErrorsServiceBaseUrl">https://oncall.keyapp.org/</string>
 
-    <string name="strigaProxyServiceBaseUrl">https://payment-rust.keyapp.org/striga/api/</string>
-    <string name="strigaProxyServiceProdBaseUrl">https://payment.key.app/striga/api/</string>
+    <string name="strigaProxyServiceBaseUrl">https://payment-rust.keyapp.org/</string>
```
gslevinkov, line 43: do you need to add prod strigaProxyServiceBaseUrl in release strings.xml?

gslevinkov (approved):

core/src/main/res/values/environment_config.xml
eduardmaximovich: it's not used now, just historically they were there together, dev and prod urls, but yes, you're right, it's better to move prod url to the `release` strings to avoid doing this later

Davran331122 (approved):

kabramovp2p (approved):

PR #2076, opened 2023-08-03
PWN-9504 - connection errors handling + check tokens on empty in SolanaTokensLoader

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9504

## Description of Work

PWN-9504 - connection errors handling + check tokens on empty in SolanaTokensLoader


app/src/main/java/org/p2p/wallet/user/interactor/TokenMetadataInteractor.kt
```diff
@@ -62,6 +82,7 @@ class TokenMetadataInteractor(
 
         if (tokensMetadata?.tokens == null) {
             Timber.tag(TAG).e("Local file not found!")
+            subscribeOnInternetConnection()
```
gslevinkov, line 85: can we do it in the splash screen? Jobs and stuff is not domain stuff

kabramovp2p: splash will die
metadata needed on Main Screen with reload

PR #2085, opened 2023-08-07
NO-TICKET - Fix onboarding metadata request

gslevinkov:


app/src/main/java/org/p2p/wallet/auth/interactor/MetadataInteractor.kt
```diff
@@ -78,25 +78,15 @@ class MetadataInteractor(
         return DeviceInfoHelper.getCurrentDeviceName() != metadata.deviceShareDeviceName
     }
 
-    private fun getEthereumPublicKey(): String? {
+    private fun getWeb3EthereumPublicKey(): String? {
         return signUpDetailsStorage.getLastSignUpUserDetails()
             ?.signUpDetails
             ?.ethereumPublicKey
-            ?: tryToGetEthAddress()
```
gslevinkov, line 85: only web3 is used, not ETH

core/src/main/java/org/p2p/core/crashlytics/helpers/CrashHttpLoggingInterceptor.kt
```diff
@@ -31,8 +31,8 @@ class CrashHttpLoggingInterceptor : Interceptor {
 
         val responseBody = response.bodyAsString()
         val responseBodySize = responseBody.length
-        if (responseBodySize < 10_000) {
-            if (!response.isSuccessful) {
+        if (responseBodySize < 7_000) {
+            if (!response.isSuccessful || responseBody.contains("error")) {
```
gslevinkov, line 35: sometimes response comes with 200, but with error inside, trying to identify it like this

eduardmaximovich (approved):

kabramovp2p (approved):

PR #2093, opened 2023-08-08
PWN-9475 - can token be hidden extension usage implementation

kabramovp2p:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/PWN-9475

## Description of Work

PWN-9475 - can token be hidden extension usage implementation

### DIFF
--- a/app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoFragment.kt
+++ b/app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoFragment.kt
@@ -109,7 +109,7 @@ class MyCryptoFragment :
     }
 
     private fun showTokenVisibilityStateChangeDialog(item: TokenCellModel) {
-        if (item.isDefinitelyHidden || item.payload.canTokenBeHidden) {
+        if (item.payload.canTokenBeHidden) {
             TokenVisibilityChangeBottomSheet.show(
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoFragment.kt
```diff
@@ -109,13 +109,15 @@ class MyCryptoFragment :
...
 
     private fun showTokenVisibilityStateChangeDialog(item: TokenCellModel) {
-        TokenVisibilityChangeBottomSheet.show(
-            fm = childFragmentManager,
-            token = item.payload,
-            isTokenHidden = item.isDefinitelyHidden,
-            requestKey = KEY_REQUEST,
-            resultKey = KEY_RESULT_TOKEN
-        )
+        if (item.isDefinitelyHidden || item.payload.canTokenBeHidden) {
```
1sakov, line 112: `item.isDefinitelyHidden` is not used for showing dialog. 
`canTokenBeHidden` should be used only

kabramovp2p: `FIXED`

gslevinkov (approved):

1sakov (approved):

PR #2137, opened 2024-01-11
ETH-670 - now stablecoins in send are first, then tokens sorted by balance

eduardmaximovich:
## Jira Ticket

https://linear.app/etherean/issue/ETH-670/[android]-key-app-pri-perehode-na-ekran-send-ne-predvybran-usdc

### DIFF
--- a/core/src/main/java/org/p2p/core/token/Token.kt
+++ b/core/src/main/java/org/p2p/core/token/Token.kt
@@ -17,6 +17,7 @@ import org.p2p.core.utils.asUsd
 import org.p2p.core.utils.formatFiat
 import org.p2p.core.utils.formatToken
 import org.p2p.core.utils.isZero
+import org.p2p.core.utils.orZero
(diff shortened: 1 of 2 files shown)


gslevinkov (approved):

core/src/main/java/org/p2p/core/token/Token.kt
```diff
@@ -269,3 +270,20 @@ fun List<Token.Active>.findByMintAddress(mintAddress: String): Token.Active? =
...
+    return sortedWith(
+        compareBy<Token.Active> {
+            when (it.tokenSymbol) {
+                USDC_SYMBOL -> 1
+                USDT_SYMBOL -> 2
+                else -> 3
+            }
+        }
+            .thenComparing(
+                compareByDescending {
```
gslevinkov, line 284: can be extracted to a val and passes here so it's more readable
same with compareBy<Token.Active> {

PR #2164, opened 2024-02-01
ETH-853, ETH-852, ETH-851 - pnl updates

eduardmaximovich:
## Jira Ticket

https://linear.app/etherean/issue/ETH-853/[android]-net-povtoryayushegosya-vyzova-servisa-raz-v-5-minut
https://linear.app/etherean/issue/ETH-852/[android]-ne-hvataet-obnovleniya-po-pull-to-refresh
https://linear.app/etherean/issue/ETH-851/[android]-nizkoe-razreshenie-u-kartinki-v-shtorke

## Description of Work

- added PNL update when pull-to-refresh is triggered
- added more rich cat images
- removed PnlStorage; now we store PNL data inside memory
- PNL interactor converted into PnlDataObserver, it works now as a poller


app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoPresenter.kt
```diff
@@ -107,6 +112,17 @@ class MyCryptoPresenter(
         }
     }
 
+    private fun observePnlData() {
+        pnlDataSubscription?.cancel()
+        pnlDataSubscription = launch {
+            pnlInteractor.state
+                .onEach {
+                    handleTokenState(tokenServiceCoordinator.observeLastState().value)
```
gslevinkov, line 120: combine?

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoPresenter.kt
eduardmaximovich: they are synchronized only at the first launch and when forcibly refreshed; thereafter new values are produced at different times

PR #2173, opened 2024-02-05
NO-TICKET - hotfix: make rentEpoch field BigInteger instead of int an…

eduardmaximovich:


buildSrc/src/main/java/AppVersions.kt
```diff
@@ -1,7 +1,7 @@
 object AppVersions {
     const val VERSION_MAJOR = 2
     const val VERSION_MINOR = 11
-    const val VERSION_HOTFIX = 1
+    const val VERSION_HOTFIX = 2
```
gslevinkov, line 4: rollback

gslevinkov (approved):

PR #2177, opened 2024-02-06
PWN-841 - Fix switching card when usd is selected

gslevinkov:


app/src/main/java/org/p2p/wallet/moonpay/ui/new/BuyPresenter.kt
```diff
@@ -230,25 +231,35 @@ class NewBuyPresenter(
...
+
+        return when {
+            isUsdSelected || isEuValidCurrencySelected -> {
+                switchToCardPaymentMethod()
+                isValidCurrencyForPay()
+            }
+            selectedCurrency == FiatCurrency.GBP && currentAlpha3Code != BANK_TRANSFER_UK_CODE -> {
+                switchToCardPaymentMethod()
+                isValidCurrencyForPay()
+            }
```
eduardmaximovich: duplication, I suggest using simple `if` instead to avoid confusions about different states, something like this:
```kotlin
if (currentPaymentMethod.method != PaymentMethod.MethodType.BANK_TRANSFER) {
    return true
}

val isUsdSelected = selectedCurrency == FiatCurrency.USD
val isEuValidCurrencySelected = currentAlpha3Code == BANK_TRANSFER_UK_CODE && selectedCurrency == FiatCurrency.EUR
val isGbpWithDifferentCode = selectedCurrency == FiatCurrency.GBP && currentAlpha3Code != BANK_TRANSFER_UK_CODE

if (isUsdSelected || isEuValidCurrencySelected || isGbpWithDifferentCode) {
    switchToCardPaymentMethod()
    return isValidCurrencyForPay()
}

return true
```

eduardmaximovich (approved):

app/src/main/java/org/p2p/wallet/moonpay/ui/new/BuyPresenter.kt
gslevinkov: yeaah, sounds good

PR #2178, opened 2024-02-06
PWN-877 - add PNL feature toggle

gslevinkov:
## Jira Ticket

https://p2pvalidator.atlassian.net/browse/

## Description of Work

## Screenshots (Optional)

### DIFF
--- a/app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoPresenter.kt
+++ b/app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoPresenter.kt
@@ -87,7 +87,7 @@ class MyCryptoPresenter(
                 // pnl must restart it's 5 minutes timer after force refresh
                 // and we should restart it only if it was initially started
                 // other cases might indicate that we didn't start observer due to empty state
-                if (pnlInteractor.isStarted()) {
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/p2p/wallet/home/ui/crypto/MyCryptoPresenter.kt
```diff
@@ -87,7 +87,7 @@ class MyCryptoPresenter(
                 // pnl must restart it's 5 minutes timer after force refresh
                 // and we should restart it only if it was initially started
                 // other cases might indicate that we didn't start observer due to empty state
-                if (pnlInteractor.isStarted()) {
+                if (!pnlInteractor.canBeStarted()) {
```
eduardmaximovich: it looks there are logical mistake: `if it can't be started - restart`
I suppose you wanted to check whether it can be started at all, considering feature toggle condition, in this case you need to add one more `if`:
```
if(pnlIntereactor.canBeStarted() && pnlInteractor.isStarted()) {
    pnlInteractor.restartAndRefresh()
}
```

checking for "isStarted" is necessary here, because we should **restart** only if it was initially started after the first tokens load (see `handleTokenState` `is UserTokensState.Loaded` block)

eduardmaximovich (approved):

PR #2196, opened 2024-02-14
PWN-945 - History and Send fixes

gslevinkov:


app/src/main/java/org/p2p/wallet/history/model/rpc/RpcHistoryTransaction.kt
```diff
@@ -59,19 +58,14 @@ sealed class RpcHistoryTransaction(
...
-            }
+        fun getTotal(): String {
+            val symbol = getSymbol(isBurn)
+            val amount = amount.total.abs().formatTokenWithSymbol(tokenSymbol, tokenDecimals)
+            return "$symbol$amount"
+        }
 
         fun getFormattedAbsTotal(): String {
-            return "${amount.total.abs().scaleLong().toPlainString()} $tokenSymbol"
+            return amount.total.abs().formatTokenWithSymbol(tokenSymbol, tokenDecimals)
```
gslevinkov: so strange that we didn't use decimals in history at all

app/src/main/java/org/p2p/wallet/history/model/rpc/RpcHistoryTransaction.kt
```diff
@@ -129,31 +124,26 @@ sealed class RpcHistoryTransaction(
         override val blockNumber: Int,
         override val status: HistoryTransactionStatus,
         override val type: RpcHistoryTransactionType,
-        val sourceAddress: String,
-        val destinationAddress: String,
+        val tokenAAddress: String,
+        val tokenBAddress: String,
```
gslevinkov: sorry, i renamed it because it was hard to understand which is which

eduardmaximovich (approved):

PR #2200, opened 2024-02-16
PWN-978 PWN-864 - navigate to Home when ref, fix set_referent request

gslevinkov:


app/src/main/java/org/p2p/wallet/common/feature_toggles/toggles/remote/ReferralProgramEnabledFeatureToggle.kt
```diff
@@ -8,4 +8,7 @@ class ReferralProgramEnabledFeatureToggle(
     override val featureKey: String = "referral_program_enabled"
     override val featureDescription: String = "Enables referral program functions"
     override val defaultValue: Boolean = false
+
+    override val value: Boolean
+        get() = true
```
gslevinkov:
```suggestion
```

app/src/main/java/org/p2p/wallet/auth/ui/pin/newcreate/NewCreatePinFragment.kt
```diff
@@ -125,6 +126,7 @@ class NewCreatePinFragment :
         } else {
             arrayListOf()
         }
+        (activity as? RootActivity)?.onSignInSuccess()
```
eduardmaximovich, line 129: I missed that, thanks! 
you can remove the next line, because onSignInSuccess in activity does all necessary navigation stuff
```
 popAndReplaceFragment(
            MainContainerFragment.create(actions),
            inclusive = true
```

eduardmaximovich: ah I see, you've changed onSignInSuccess

eduardmaximovich (approved):

PR #2204, opened 2024-02-19
ETH-975 - fixed use max when fee payer is the source token

eduardmaximovich:
## Jira Ticket

https://linear.app/etherean/issue/ETH-975/[android]-esli-token-kompensacii-i-otpravki-sovpadaet-to-pri

## Description of Work

- fixed calculating rent exemption for token2022
- fixed send slider state evaluation when "max" is used

### DIFF
--- a/app/src/main/java/org/p2p/wallet/send/SendFeeRelayerManager.kt
+++ b/app/src/main/java/org/p2p/wallet/send/SendFeeRelayerManager.kt
@@ -6,28 +6,21 @@ import java.math.BigInteger
 import kotlin.properties.Delegates.observable
 import kotlinx.coroutines.CancellationException
 import org.p2p.core.token.Token
-import org.p2p.core.utils.Constants
 import org.p2p.core.utils.formatTokenWithSymbol
 import org.p2p.core.utils.fromLamports
-import org.p2p.core.utils.isZero
 import org.p2p.core.utils.toLamports
 import org.p2p.core.utils.toUsd
-import org.p2p.solanaj.core.FeeAmount
 import org.p2p.solanaj.kits.AccountInfoTokenExtensionConfig.Companion.interestBearingConfig
 import org.p2p.solanaj.kits.AccountInfoTokenExtensionConfig.Companion.transferFeeConfig
 import org.p2p.solanaj.kits.TokenExtensionsMap
-import org.p2p.solanaj.programs.TokenProgram
 import org.p2p.solanaj.rpc.RpcSolanaRepository
-import org.p2p.wallet.feerelayer.interactor.FeeRelayerTopUpInteractor
 import org.p2p.wallet.feerelayer.model.FeeCalculationState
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/p2p/wallet/send/SendFeeRelayerManager.kt
```diff
@@ -81,6 +68,14 @@ class SendFeeRelayerManager(
 
     private val alternativeTokensMap: HashMap<String, List<Token.Active>> = HashMap()
 
+    private val TokenExtensionsMap.actualTransferFeePercent: BigDecimal
+        get() {
+            return this.transferFeeConfig
+                ?.getActualTransferFee(currentSolanaEpoch)
+                ?.transferFeePercent
```
gslevinkov, line 76: orZero?

app/src/main/java/org/p2p/wallet/send/interactor/usecase/CalculateSendFeesUseCase.kt
```diff
@@ -0,0 +1,194 @@
...
+                fee = FeeRelayerFee(
+                    feesInSol = feesInSol,
+                    feesInFeePayerToken = feesInFeePayerToken,
+                    feesInSourceToken = feesInSourceToken,
+                    expectedFee = feesInSol
+                )
+            )
+        } catch (e: CancellationException) {
+            Timber.i("Fee calculation cancelled")
+            return FeeCalculationState.Cancelled
```
gslevinkov, line 81: multiple returns instead of one root return

app/src/main/java/org/p2p/wallet/send/interactor/usecase/CalculateSendFeesUseCase.kt
```diff
@@ -0,0 +1,194 @@
...
+     *
+     * todo: it looks like we don't have correct prices for tokens at a moment
+     *
+     * @return zero if tokens are different, or min rent exemption in source token
+     */
+    private suspend fun getSendServiceMinRentExemptionForSameTokens(
+        sourceToken: Token.Active,
+        feePayerToken: Token.Active
+    ): BigInteger {
+        return if (!sourceToken.isSOL && sourceToken.mintAddress == feePayerToken.mintAddress) {
```
gslevinkov: early if? 

if (sourceToken.isSol) return
if (sourceToken.mintAddress != feePayerToken.mintAddress) return

gslevinkov (approved):

app/src/main/java/org/p2p/wallet/send/interactor/usecase/CalculateSendFeesUseCase.kt
eduardmaximovich: early return allows to avoid evaluating all the following conditions, but ok, I'll move zero check into `when`

PR #2219, opened 2024-02-26
ETH-1059, ETH-1068 - cherry-picks to release; bump app build version

eduardmaximovich:
cherry-picks + version bump


buildSrc/src/main/java/AppVersions.kt
```diff
@@ -3,5 +3,5 @@ object AppVersions {
     const val VERSION_MINOR = 12
     const val VERSION_HOTFIX = 0
     const val VERSION_PATCH = 0
```
gslevinkov:
```suggestion
    const val VERSION_PATCH = 1
```

